PR #9, opened 2015-05-16
Optimized OgnlContext constructors, reduced collection sizes

danielfernandez:
I've optimized the `OgnlContext` constructors so that:

   * The `_values` map object is not instantiated if not really needed. This map (which was a large one, 23 entries) was being always created, even if the `OgnlContext` constructor being used immediately replaced it by one of its arguments. I'm executing some thousands of OGNL expressions per second and I'm always using `new OgnlContext(myOwnMap)`, so this meant a lot of spurious, large `HashMap` objects being created without any need and a waste in memory usage.

   * The `_typeStack` and `_accessorStack` objects, of type `List`, were being initialized to their default size which is `10`, but very few OGNL expressions actually use 10 stack positions. Normally they don't go over `2` or `3`. So by reducing the initial size of those collections to `3`, a noticeable save in memory could be obtained under heavy loads.

### DIFF
--- a/src/java/ognl/OgnlContext.java
+++ b/src/java/ognl/OgnlContext.java
@@ -72,7 +72,7 @@ public class OgnlContext extends Object implements Map
     private Evaluation _lastEvaluation;
     private boolean _keepLastEvaluation = DEFAULT_KEEP_LAST_EVALUATION;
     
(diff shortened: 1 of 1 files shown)


lukaszlenart:
Looks good, let's roll!

PR #10, opened 2015-05-16
Optimized the use of java.lang.Class#getInterfaces() in OgnlCompiler

danielfernandez:
Most of my profiling sessions show `java.lang.Class#getInterfaces()` as one of the *hot spots* under heavy OGNL load, because of being used at several methods in `ognl.enhance.ExpressionCompiler`, especially `ognl.enhance.ExpressionCompiler#getInterfaceClass(java.lang.Class)`

Indeed it seems this `java.lang.Class#getInterfaces()` method, even if being native, is quite slow for some strange reason. I've read a couple of complaints about it also related with Hibernate (see [1] and [2]).

Fortunately, I was able to importantly reduce the amount of times this method has to be called in `ExpressionCompiler` without losing any functionality, just re-structuring a bit the code. This should save a noticeable amount of execution time.

[1] https://forum.hibernate.org/viewtopic.php?p=2404651
[2] https://jira.atlassian.com/browse/CRUC-4518

### DIFF
--- a/src/java/ognl/enhance/ExpressionCompiler.java
+++ b/src/java/ognl/enhance/ExpressionCompiler.java
@@ -202,8 +202,11 @@ public class ExpressionCompiler implements OgnlExpressionCompiler {
         if (Modifier.isPublic(clazz.getModifiers()) && clazz.isInterface())
             return clazz.getName();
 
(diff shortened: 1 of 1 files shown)


lukaszlenart:
Very neat!

PR #11, opened 2015-05-16
Optimization: avoid using java.lang.String#replaceAll() in most cases

danielfernandez:
Several parts of OGNL call the `java.lang.String#replaceAll(...)` method, which is extremely inefficient because it first compiles a `java.util.regex.Pattern`, then a `Matcher`, and finally asks this `Matcher` to perform the replace operation.

In order to avoid most of the calls to this method without disrupting the code base too much, I've added before each call a condition based on a call to `java.lang.String#indexOf(...)`, which is hugely more efficient because it acts directly on the `String`'s inner `char[]`. And only if this `indexOf(...)` determines there is a possibility that the `replaceAll(...)` operation would actually *do something*, `replaceAll(...)` will be called.


lukaszlenart:
:+1:

Issue #16, opened 2016-03-08
Unit tests fail when using Java 8 to build maven package


marvkis:
Hi,

When doing a "mvn test" with Java 8 the tests fail with these error messages:

```
Results :

Failed tests:
  messages.format('ShowAllCount', one) (foo)(org.ognl.test.MethodTest)

Tests in error:
  {'1','2','3'}(org.ognl.test.ArrayElementsTest)
  { true, !false }(org.ognl.test.ArrayElementsTest)
  {stringValue, getMap()}(org.ognl.test.PropertyTest)
  {'stringValue', map["test"].map["size"]}(org.ognl.test.PropertyTest)
  test_Complicated_List(org.ognl.test.ASTPropertyTest)
  null in {true,false,null} (true)(org.ognl.test.ArithmeticAndLogicalOperatorsTest)
  null not in {true,false,null} (false)(org.ognl.test.ArithmeticAndLogicalOperatorsTest)
  5 in {true,false,null} (false)(org.ognl.test.ArithmeticAndLogicalOperatorsTest)
  5 not in {true,false,null} (true)(org.ognl.test.ArithmeticAndLogicalOperatorsTest)
  #y + "1" (11)(org.ognl.test.ArithmeticAndLogicalOperatorsTest)
  "1" + #y (11)(org.ognl.test.ArithmeticAndLogicalOperatorsTest)
  { false, true, null, 0, 1. } ([false, true, null, 0, 1.0])(org.ognl.test.ConstantTest)
  messages.format('ShowAllCount', {one}) (foo)(org.ognl.test.MethodTest)
  messages.format('ShowAllCount', {one, two}) (foo)(org.ognl.test.MethodTest)
  {theInt + 1}(org.ognl.test.PropertyArithmeticAndLogicalOperatorsTest)

Tests run: 730, Failures: 1, Errors: 15, Skipped: 0
````
In the logs we see a lot of messages like:
```
Running org.ognl.test.ArrayElementsTest
Caught exception java.lang.RuntimeException: java.io.IOException: invalid constant type: 18
Caught exception java.lang.RuntimeException: java.io.IOException: invalid constant type: 18
Tests run: 11, Failures: 0, Errors: 2, Skipped: 0, Time elapsed: 0.026 sec <<< FAILURE!
```
The invalid constant type exception referes to Java 8 classes. Okay, OGNL uses javassist to access the java objects. The utilized version is 3.11, which dosn't support Java 8. 
[...]

Issue #17, opened 2016-03-08
ArrayIndexOutOfBoundsException when trying to access BeanFactory


marvkis:
Hi,

when trying to access a BeanFactory with `factory.getBean("test")` an ArrayIndexOutOfBoundsExceptions is thrown:

```java
java.lang.ArrayIndexOutOfBoundsException
	at java.lang.System.arraycopy(Native Method)
	at ognl.OgnlRuntime.callAppropriateMethod(OgnlRuntime.java:1265)
	at ognl.ObjectMethodAccessor.callMethod(ObjectMethodAccessor.java:68)
	at ognl.OgnlRuntime.callMethod(OgnlRuntime.java:1370)
	at ognl.ASTMethod.getValueBody(ASTMethod.java:90)
	at ognl.SimpleNode.evaluateGetValueBody(SimpleNode.java:212)
	at ognl.SimpleNode.getValue(SimpleNode.java:258)
	at ognl.ASTChain.getValueBody(ASTChain.java:141)
	at ognl.SimpleNode.evaluateGetValueBody(SimpleNode.java:212)
	at ognl.SimpleNode.getValue(SimpleNode.java:258)
	at ognl.ASTMethod.getValueBody(ASTMethod.java:87)
...
```

I created a small Dummy-BeanFactory to reproduce the issue:

```java
package org.ognl.test.objects;

import java.util.Arrays;

public class BeanFactory {
	//---------------------------------------------------------------------
	// Implementation of BeanFactory interface
	//---------------------------------------------------------------------

	public Object getBean(String name) {
		return "NamedBean: "+name;
	}

	public <T> T getBean(String name, Class<T> requiredType) {
		return (T) ("NamedTypedBean: "+name+" "+requiredType.getSimpleName());
	}

	public <T> T getBean(Class<T> requiredType) {
		return (T) ("TypedBean: "+requiredType.getSimpleName());
	}

	public Object getBean(String name, Object... args) {
		return "NamedBeanWithArgs: "+name+" "+Arrays.toString(args);
	}

}
```

.. might also be related to the issue in #16

PR #19, opened 2016-03-10
This is the initial version of an 'java like' method matching for OGNL.

marvkis:
This fixes:
#16 
#17 
https://issues.apache.org/jira/browse/OGNL-250

and maybe even more issues related to wrong methods. 

But it's a huge change in OGNL method matching so it also might break some things.

All unit tests pass for Java 7 and Java 8.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -66,9 +66,9 @@
             <scope>test</scope>
         </dependency>
         <dependency>
-            <groupId>javassist</groupId>
+            <groupId>org.javassist</groupId>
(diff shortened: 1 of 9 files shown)


lukaszlenart:
I have built it and tested with the latest Struts version (2.5) - looks like all tests passed but something potentially can explode latter on ;-) But it may also solve that problem http://struts.apache.org/docs/security.html#Security-OGNLisusedtocallaction'smethods

lukaszlenart:
Let's merge and release a new version!

PR #20, opened 2016-03-25
Update OgnlRuntime.java

249043822:
this position is not comfortable with version 3.0.6, for example, javabean field name is sL, set method should be setSL(), if use version 3.0.13,  set method return setsL(), it's wrong,  then sL cann't be set value


lukaszlenart:
I assume that this PR can be closed as #21 was opened?

PR #21, opened 2016-03-25
Update OgnlRuntime.java

249043822:
this position is not comfortable with version 3.0.6, for example, javabean field name is sL, set method should be setSL(), if use version 3.0.13,  set method return setsL(), it's wrong,  then sL cann't be set value

### DIFF
--- a/src/java/ognl/OgnlRuntime.java
+++ b/src/java/ognl/OgnlRuntime.java
@@ -1849,15 +1849,11 @@ public class OgnlRuntime {
                 return propertyName;
             }
         }
-        char first = propertyName.charAt(0);
-        char second = propertyName.charAt(1);
-        if (Character.isLowerCase(first) && Character.isUpperCase(second)) {
-            return propertyName;
-        } else {
-            char[] chars = propertyName.toCharArray();
-            chars[0] = Character.toUpperCase(chars[0]);
-            return new String(chars);
-        }
+
+        char[] chars = propertyName.toCharArray();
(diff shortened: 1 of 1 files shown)


lukaszlenart:
Current version is using proper capitalisation logic which aligns with [JavaBeans Specification](http://download.oracle.com/otndocs/jcp/7224-javabeans-1.01-fr-spec-oth-JSpec/). You can perform simple test with your IDE, define a field `private long sL` and use IDE to generate setter and getter for that field and you will get:
```java
	private long sL;

	public long getsL() {
		return sL;
	}

	public void setsL(long sL) {
		this.sL = sL;
	}
```

Here you have a dedicated quote from the spec:
> 8.8 Capitalization of inferred names.
> When we use design patterns to infer a property or event name, we need to decide what rules to follow for capitalizing the inferred name. If we extract the name from the middle of a normal mixedCase style Java name then the name will, by default, begin with a capital letter.

> Java programmers are accustomed to having normal identifiers start with lower case letters. Vigorous reviewer input has convinced us that we should follow this same conventional rule for property and event names.

> Thus when we extract a property or event name from the middle of an existing Java name, we normally convert the first character to lower case. However to support the occasional use of all upper-case names, we check if the first two characters of the name are both upper case and if so leave it alone. So for example,
> “FooBah” becomes “fooBah”
> “Z” becomes “z”
> “URL” becomes “URL”

> We provide a method Introspector.decapitalize which implements this conversion rule.

Issue #23, opened 2016-05-06
Exception selecting overloaded method in 3.1.4


danielfernandez:
In Thymeleaf there is an `Aggregates` class which contains an overloaded `avg(...)` method. See the specific class [here](https://github.com/thymeleaf/thymeleaf/blob/a4d3753dc3bbafdbeeab19a36f33c27daf313946/src/main/java/org/thymeleaf/expression/Aggregates.java#L181).

All the different variants of this method look like this:

```java
    public BigDecimal avg(final Iterable<? extends Number> target) {...}
    public BigDecimal avg(final Number[] target) {...}
    public BigDecimal avg(final byte[] target) {...}
    public BigDecimal avg(final short[] target) {...}
    public BigDecimal avg(final int[] target) {...}
    public BigDecimal avg(final long[] target) {...}    
    public BigDecimal avg(final float[] target) {...}
    public BigDecimal avg(final double[] target) {...}
```

In one of my tests, I create an object using an OGNL expression, like:
```
nums01 = { 5, 5 }
```
Which creates a `java.util.ArrayList<java.lang.Integer>` for the `nums01` variable.

Then I use this variable as an argument to an `avg` call in an OGNL expression such as:
```
#aggregates.avg(nums01)
```

This worked perfectly until OGNL 3.1.3, and the first version of the overloaded method (the one receiving an `Iterable<? extends Number>`) was executed. However, in OGNL 3.1.4 I receive this exception:
```
java.lang.IllegalArgumentException: Can't decide wich method to use: "public java.math.BigDecimal org.thymeleaf.expression.Aggregates.avg(float[])" or "public java.math.BigDecimal org.thymeleaf.expression.Aggregates.avg(long[])"
	at ognl.OgnlRuntime.findBestMethod(OgnlRuntime.java:1427)
	at ognl.OgnlRuntime.getAppropriateMethod(OgnlRuntime.java:1283)
	at ognl.OgnlRuntime.callAppropriateMethod(OgnlRuntime.java:1452)
	at ognl.ObjectMethodAccessor.callMethod(ObjectMethodAccessor.java:68)
	at ognl.OgnlRuntime.callMethod(OgnlRuntime.java:1598)
	at ognl.ASTMethod.getValueBody(ASTMethod.java:91)
	at ognl.SimpleNode.evaluateGetValueBody(SimpleNode.java:212)
	at ognl.SimpleNode.getValue(SimpleNode.java:258)
[...]

PR #24, opened 2016-05-17
Issue 23

sa-ChristianNiessner:
Bugfix for https://github.com/jkuhnert/ognl/issues/23 - Exception selecting overloaded method in 3.1.4

- Improved method matching: Lists default to List<Object> because of "Java Generics Type Erasure"
- Don't bail out (with exception) immedeately if we get two methods with same score - a method with a  better score might follow. Just remember the Exception and if we really don't get an better option throw the exception


lukaszlenart:
Struts build passed with this change, I'm waiting on @danielfernandez and will merge it and push a new release. Awesome work!

lukaszlenart:
Closes #23

Issue #25, opened 2016-05-17
TypeConverter / OgnlOps 'double boxing'


sa-ChristianNiessner:
Hi,

I just pushed a new branch that also addresses another 'inconsistency' issue I was fighting in the past:
The automatic Type-Conversion puts a List (or Collection) into an array by doing new Object[] { <listclass> } instead of using the items of the list for the array. I think this would be the more desired behaviour. I also fixed the related unit tests from #16 / #19 respectively and added some new tests.

The commit [410dee5f64410f98ad4171150dc19869abd7de4c](https://github.com/secadm/ognl/commit/410dee5f64410f98ad4171150dc19869abd7de4c] (branch [TypeConverterDoubleArrayBoxing](https://github.com/secadm/ognl/commits/TypeConverterDoubleArrayBoxing)) addresses the issue.

This patch is based on #23 / #24 and so i'll create a pull request after #24 ist merged or the pull request will also provide #24. But maybe both requests can be reviewed together.

Bye,
    Chris

lukaszlenart:
Great! I would rather push a new version based on #24 and then merge this one (and push new release), it's a safer path :)

lukaszlenart:
@sa-ChristianNiessner you are free to go with this :)

sa-ChristianNiessner:
And here it is. Just got the mails about the 3.1.5 release ;)

lukaszlenart:
I see no objections, merge party!

lukaszlenart:
Fixed by merging #26

PR #26, opened 2016-05-18
Pull-Request for Isuse #25 - TypeConverter / OgnlOps 'double boxing' 

sa-ChristianNiessner:
Fixes the behaviour described in #25

### DIFF
--- a/src/test/java/org/ognl/test/MethodTest.java
+++ b/src/test/java/org/ognl/test/MethodTest.java
@@ -65,16 +65,20 @@ public class MethodTest extends OgnlTestCase
             { "testMethods.getBean('TestBean')", ROOT.getTestMethods().getBean("TestBean") } ,
             // https://issues.apache.org/jira/browse/OGNL-250 -  OnglRuntime getMethodValue fails to find method matching propertyName
             { "testMethods.testProperty", ROOT.getTestMethods().testProperty() } ,
-            // TODO: some further java <> OGNL inconsistencies: - related to https://github.com/jkuhnert/ognl/issues/16
-            //   Java 'ROOT.getTestMethods().argsTest1(Arrays.asList( ROOT.getOne() )' doesn't compile:
-            //		--> The method argsTest1(Object[]) in the type MethodTestMethods is not applicable for the arguments (List<Integer>)
-            { "testMethods.argsTest1({one})", "Array: [[1]]" }, // "Array: [[1]]" means dual-cast is done.
-            //   Java: ROOT.getTestMethods().argsTest(Arrays.asList( ROOT.getOne() ))
-            //		--> The method argsTest2(List<Object>) in the type MethodTestMethods is not applicable for the arguments (List<Integer>)
-            { "testMethods.argsTest2({one})", "List: [1]" },
+            { "testMethods.argsTest1({one})", ROOT.getTestMethods().argsTest1(Arrays.asList( ROOT.getOne() ).toArray()) },	// toArray() is automatically done by OGNL type conversion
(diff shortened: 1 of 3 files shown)


lukaszlenart:
Just pushed snapshot to Sonatype

lukaszlenart:
w00t! Struts build passed ;-)

src/test/java/org/ognl/test/MethodTest.java
```diff
@@ -65,16 +65,20 @@
...
             //   Java 'ROOT.getTestMethods().argsTest1(Arrays.asList( ROOT.getOne() )' doesn't compile:
             //		--> The method argsTest(Object[]) in the type MethodTestMethods is not applicable for the arguments (List<Integer>)
             { "testMethods.argsTest3({one})", "List: [1]" },
+            { "testMethods.showList(testMethods.getObjectList())", ROOT.getTestMethods().showList(ROOT.getTestMethods().getObjectList().toArray()) },
+            { "testMethods.showList(testMethods.getStringList())", ROOT.getTestMethods().showList(ROOT.getTestMethods().getStringList().toArray()) },
+            { "testMethods.showList(testMethods.getStringArray())", ROOT.getTestMethods().showList(ROOT.getTestMethods().getStringArray()) },
+            // TODO This one doesn't work - even 'toArray(new String[0]) returns Object[] and so the wrong method is called - currently no idea how to handle this...
+            // { "testMethods.showList(testMethods.getStringList().toArray(new String[0]))", ROOT.getTestMethods().showList(ROOT.getTestMethods().getStringList().toArray(new String[0])) },
```
lukaszlenart: This test still fails :(

sa-ChristianNiessner: Amazing - can you post me the error details? The tests pass for me and also pass on [travis-c:](https://travis-ci.org/jkuhnert/ognl/builds/131047765#L1530) What java version are you using?

lukaszlenart: I meant, that one test which is commented out now. I uncommented it and tested locally but it fails :(

sa-ChristianNiessner: Yes, this is still an issue. I think the problem is also based on "Type Erasure". It works for 'interpretation' mode (I call it Interpretation when you run an OGNL expression without compile) because then the toArray() returns a real object which is a String[] array. But for OGNL.compile Collection.toArray(String[]) would return Object[]  due to "Type Erasure" - so javaassist picks showList(Object []) instead of showList(String[]) during code generation.
The only idea I currently have is to write something like a 'return type overwriter' for Collection.toArray(<type array>) - that does an appropate cast for the object return in Collection.toArray().. Maybe you have a better idea?

lukaszlenart: Let's leave it as is... it's an abstract problem, never have had it before ;-)

lukaszlenart:
@danielfernandez do you want to test it before I will push a new release?

src/test/java/org/ognl/test/MethodTest.java
sa-ChristianNiessner: Maybe it's a thing to keep in mind for OGNL-4. I think doing this in OGNL-3 would be hard... For now the API's / method definitions on 'OGNL-Intended' classes should be clear defined then it works.

danielfernandez:
All tests work with the latest snapshot in the Thymeleaf side :) Thanks

lukaszlenart:
uff... just pushed out a new version ;-)

Issue #29, opened 2016-09-20
Ognl.GetMethods does not return default methods from interfaces implemented by parent class


vlastimil-dolejs:
Same problem as here: https://issues.apache.org/jira/browse/OGNL-249
But fix for issue 249 takes into account only interfaces directly implemented by a class and not an interfaces implemented by a parent classes.
I will prepare a pull request with a test case (and maybe a fix also)

lukaszlenart:
PR #30 merged

PR #30, opened 2016-09-20
fix for issue #29 - Ognl.GetMethods does not return default methods from interfaces implemented by parent class

vlastimil-dolejs:


lukaszlenart:
Looks good, LGTM! 👍

PR #33, opened 2016-11-14
Add access for property which read method is java8 default method

quaff:
Workaround for https://bugs.openjdk.java.net/browse/JDK-8071693

### DIFF
--- a/src/java/ognl/OgnlRuntime.java
+++ b/src/java/ognl/OgnlRuntime.java
@@ -1790,7 +1790,7 @@ public class OgnlRuntime {
                     {
                         toExamined.add(c);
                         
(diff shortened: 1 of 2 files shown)


lukaszlenart:
hmm... I wonder if this is a good approach, resolving bugs in JVM by such workarounds. After some time, even when the JVM get fixed, no one will remember about this fix.

quaff:
This issue have been opened for a long time and no response, I guess they don't think it's really bug and will not going to fix it, third-party library should ensure compatibility, because even JDK fix it in future, but affected version is widely used. 
Freemarker will fix it, https://issues.apache.org/jira/browse/FREEMARKER-24

quaff:
Ping

PR #36, opened 2017-07-21
WW-4817 Makes better decisions on methods first call

yasserzamani:
if check permission and run synchronized on method first call in order to fix [WW-4817](https://issues.apache.org/jira/browse/WW-4817)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -10,7 +10,7 @@
     <groupId>ognl</groupId>
     <artifactId>ognl</artifactId>
     <packaging>jar</packaging>
-    <version>3.1.14</version>
+    <version>3.1.15-SNAPSHOT</version>
     <name>OGNL - Object Graph Navigation Library</name>
     <description>OGNL - Object Graph Navigation Library</description>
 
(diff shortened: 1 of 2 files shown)


lukaszlenart:
Tested locally and looks good, didn't notice any problems :)

yasserzamani:
Thanks for your test. Yes, I think these changes do no have any functional test in point of view of outside of the `invokeMethod` method. Previously it got decision then cached, now it caches then decides :)

Just one more thing that I think maybe we do not have to call `_securityManager.checkPermission(getPermission(method));` for each method call and the first one for caching is enough but I decided to keep it unchanged logically as I was not sure.

lukaszlenart:
Please open PR against 3.0.x branch as well and then please open it again against the `master` branch and then you can improve the `_securityManager.checkPermission(getPermission(method));` call :)

yasserzamani:
Sure :) Do you mean that calling `_securityManager.checkPermission(getPermission(method));` in each method call is not necessary (should be removed from 3.0.x, 3.1.x and master) and the first one (i.e. in method first call and filling cache) is enough?

lukaszlenart:
No, I meant to keep changes as minimal as possible in 3.1.x and 3.0.x - so your current approach is fine. But in your third PR against the `master` branch you can optimize this call just to first method call - the `master` branch is still an experimental and a bit unstable and it targets 3.2.x series.

PR #40, opened 2017-08-26
When scanning super interfaces in OgnlRuntime#getMethods(Class, bool…

### LINKED ISSUE #35: Output in System.err 'Two methods with same method signature but not providing classes assignable?'
Here is a repro:

```java
  @Test
  public void test() throws Exception {
    String expressiosn = "list.isEmpty()";
    Object expr = Ognl.parseExpression(expressiosn);
    OgnlContext ctx = new OgnlContext();
    List<String> list = Collections.singletonList("string");
    ctx.put("list", list);
    Object value = Ognl.getValue(expr, ctx);
    Assert.assertFalse((Boolean)value);
  }
```

The above test passes, but the following message is output in System.err.

> Two methods with same method signature but not providing classes assignable? "public abstract boolean java.util.List.isEmpty()" and "public boolean java.util.AbstractCollection.isEmpty()" please report!

- The line that outputs the message is [here](https://github.com/jkuhnert/ognl/blob/7f6ebaf06b0e00630902b95ed18f9fccde27f0d9/src/java/ognl/OgnlRuntime.java#L1393).
- This seems to be a side effect of ca37319fb74ad8468acc78db5d3ba9d7b4863765 .

This was reported on [MyBatis tracker](https://github.com/mybatis/mybatis-3/issues/929) a while ago.

harawata:
…ean), only default methods should be collected instead of all declared methods. This should fix #35 .

Also, it now searches super interfaces recursively to handle deeper interface hierarchy as follows.

```java
interface InterfaceWithDefaults {
  default public void defaultMethod() { }
}

interface SubInterfaceWithDefaults extends InterfaceWithDefaults {
}

class ClassWithDefaults implements SubInterfaceWithDefaults {
}
```

```java
@Test
public void testDefaultMethodOnClass() {
  List result = OgnlRuntime.getMethods(
    ClassWithDefaults.class, "defaultMethod", false);
  assertNotNull(defaultMethod);
}
```

To achieve the recursive search, I extracted some methods. Was it OK ?
I ask because the original code seemed to avoid it intentionally for some reason.

A few not-directly-related things I noticed:

- PR #30 was not applied to `getAllMethods(Class, boolean)`.
- PR #33 does not search interfaces recursively.

If this PR is acceptable, I will send another PR to backport the change to 3.1 (and 3.0 if it's appropriate).


lukaszlenart:
Looks good to me ... give me some time to take another shot (just returned from vacations ;-)

harawata:
Hope you recharged your batteries! =)
Please take your time. I'm not surprised if I missed something.

lukaszlenart:
@harawata could you resolve that conflict? Or you can target 3.1.x

harawata:
@lukaszlenart Done!
I'll open another PR for 3.1.x.

Thanks!

lukaszlenart:
LGTM 💯  & thank you again :)

PR #41, opened 2017-11-15
Backporting #40 to ognl-3-1-x branch.

harawata:
When scanning super interfaces in `OgnlRuntime#getMethods(Class, boolean)`, only default methods should be collected instead of all declared methods.
Also, super interfaces should be searched recursively so that it can handle deeper interface hierarchy.

### DIFF
--- a/src/java/ognl/OgnlRuntime.java
+++ b/src/java/ognl/OgnlRuntime.java
@@ -1789,43 +1789,57 @@ public class OgnlRuntime {
                 if ((result = (Map) cache.get(targetClass)) == null)
                 {
                     result = new HashMap(23);
-
-                    List<Class> toExamined = new LinkedList<Class>();
-                    for (Class c = targetClass; c != null; c = c.getSuperclass())
-                    {
-                        toExamined.add(c);
-                        
-                        // Including interfaces is needed as from Java 8 intefaces can implement defaul methods
-                        toExamined.addAll(Arrays.asList(c.getInterfaces()));
-                    }
-
-                    for (Class c : toExamined)
-                    {
-                        Method[] ma = c.getDeclaredMethods();
-
-                        for (int i = 0, icount = ma.length; i < icount; i++)
-                        {
-                            // skip over synthetic methods
-
-                            if (!isMethodCallable(ma[i]))
-                                continue;
-
-                            if (Modifier.isStatic(ma[i].getModifiers()) == staticMethods)
-                            {
-                                List ml = (List) result.get(ma[i].getName());
-
-                                if (ml == null)
-                                    result.put(ma[i].getName(), ml = new ArrayList());
-
-                                ml.add(ma[i]);
-                            }
-                        }
-                    }
+                    collectMethods(targetClass, result, staticMethods);
                     cache.put(targetClass, result);
                 }
             }
         }
         return result;
+                    }
+
+    private static void collectMethods(Class c, Map result, boolean staticMethods) {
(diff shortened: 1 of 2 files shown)


lukaszlenart:
LGTM 💯  & thank you :)

lukaszlenart:
OGNL 3.1.16 is out, could you test it and report any problems?

lukaszlenart:
Tested with Struts 2.6-SNAPSHOT on Java 8 and works :)

harawata:
Thank you, @lukaszlenart !
I'll test it later tonight.

harawata:
@lukaszlenart 
I have tested 3.1.16 with MyBatis and it seems to be working fine. Thanks again!

lukaszlenart:
@harawata osm! Thank you so much for your contribution :)

YSavanier:
Hello, I'm sorry but I think the PR was missing important diffs since code added by harawata ( I'm a big fan of mybatipse by the way thank you for your hard work :D ) on class OgnlRuntime [line 2151](https://github.com/harawata/ognl/blob/ad6cda80bed6ca080b18f58b45d5bdcec12d5cac/src/java/ognl/OgnlRuntime.java#L2151) to the getDeclaredMethods method included an interface default methods scanning routine which is missing of current master branch hence breaking unit Java8Test testGetDeclaredMethods as reported in [issue 51](https://github.com/jkuhnert/ognl/issues/51)

Could you please consider adding the mssing code in next minor release please as for today we were forced to ovveride the OgnlRuntime class to add the missing code :+1: 

`                    // Add default methods on interface
                    for (Class c = targetClass; c != null; c = c.getSuperclass()) {
                    	for(Class intf : c.getInterfaces()){
	                        Method[] methods = intf.getDeclaredMethods();
	
	                        for (int i = 0; i < methods.length; i++) {
	                            if (Modifier.isAbstract(methods[i].getModifiers()))
	                                continue;
	
	                            String ms = methods[i].getName();
	
	                            if (ms.endsWith(baseName)) {
	                                boolean isSet = false, isIs = false;
	
	                                if ((isSet = ms.startsWith(SET_PREFIX)) || ms.startsWith(GET_PREFIX)
	                                    || (isIs = ms.startsWith(IS_PREFIX))) {
	                                    int prefixLength = (isIs ? 2 : 3);
	
	                                    if (isSet == findSets) {
	                                        if (baseName.length() == (ms.length() - prefixLength)) {
	                                            if (result == null) {
	                                                result = new ArrayList();
	                                            }
	                                            result.add(methods[i]);
	                                        }
	                                    }
	                                }
	                            }
	                        }
                    	}
                    }`

Thank you very much to both of you.

harawata:
Hi @YSavanier ,

That is one of the 'not-directly-related things' that I mentioned in #40 .
I ignored it because it didn't affect MyBatis :P
Anyway, I'll see what I can do.

And I'm glad you found the plugin useful! :)

YSavanier:
Yeah I didn't encountered the problem in MyBatis (for which by the way I'm also glad you took the lead after eduardo macaron took a break from it), I was using the engine on a JOpenDocument stack to generate documents using embeded expression language so it used another path to fill the getmethodscache map.

YSavanier:
Yeah I didn't encountered the problem in MyBatis (for which by the way I'm also glad you took the lead after eduardo macaron took a break from it), I was using the engine on a JOpenDocument stack to generate documents using embeded expression language so it used another path to fill the cacheGetMethod map.

PR #46, opened 2018-03-18
Support concurrent on DefaultClassResolver

kazuki43zoo:
The `DefaultClassResolver` is not thread-safe.
I will propose to support concurrent using `ConcurrentHashMap` on the `DefaultClassResolver`.

WDYT?

### DIFF
--- a/src/main/java/ognl/DefaultClassResolver.java
+++ b/src/main/java/ognl/DefaultClassResolver.java
@@ -49,23 +49,23 @@ public class DefaultClassResolver extends Object implements ClassResolver
         super();
     }
 
-	public Class classForName(String className, Map context) throws ClassNotFoundException
-	{
-	    Class       result = null;
+    public Class classForName(String className, Map context) throws ClassNotFoundException
+    {
+        Class       result = null;
 
         if ((result = (Class)classes.get(className)) == null) {
-    		try {
-    		    result = Class.forName(className);
-    		} catch (ClassNotFoundException ex) {
-    			if (className.indexOf('.') == -1) {
-    			    result = Class.forName("java.lang." + className);
-        			classes.put("java.lang." + className, result);
-        		}
-    		}
-            if (result != null) {
-				classes.put(className, result);
+            try {
+                result = Class.forName(className);
+            } catch (ClassNotFoundException ex) {
+                if (className.indexOf('.') == -1) {
+                    result = Class.forName("java.lang." + className);
+                    classes.put("java.lang." + className, result);
+                }
             }
-    	}
-	    return result;
-	}
+            if (result != null) {
+                classes.put(className, result);
+            }
+        }
+        return result;
+    }
 }
--- a/src/main/java/ognl/DefaultClassResolver.java
+++ b/src/main/java/ognl/DefaultClassResolver.java
@@ -42,7 +42,7 @@ import java.util.concurrent.ConcurrentHashMap;
  */
 public class DefaultClassResolver extends Object implements ClassResolver
 {
-    private Map     classes = new ConcurrentHashMap(101);
+    private final Map<String, Class> classes = new ConcurrentHashMap<>(101);
 
     public DefaultClassResolver()
     {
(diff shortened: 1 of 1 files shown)


src/main/java/ognl/DefaultClassResolver.java
```diff
@@ -41,28 +42,30 @@
  */
 public class DefaultClassResolver extends Object implements ClassResolver
 {
-    private Map     classes = new HashMap(101);
+    private Map     classes = new ConcurrentHashMap(101);
```
lukaszlenart: Can you use generics here? This will reduce casting below.

src/main/java/ognl/DefaultClassResolver.java
```diff
@@ -41,28 +42,30 @@
...
-        		}
-    		}
-			classes.put(className, result);
-    	}
-	    return result;
-	}
+            try {
+                result = Class.forName(className);
+            } catch (ClassNotFoundException ex) {
+                if (className.indexOf('.') == -1) {
```
lukaszlenart: I think when we are here, this `if` clause can be moved out of the catch and implemented as follow (`classForName` already throws `ClassNotFoundException`):

```java
if (className.indexOf('.') == -1) {
    result = Class.forName("java.lang." + className);
    classes.put("java.lang." + className, result);
} else {
    classes.put(className, Class.forName(className));
}
```

then `result` and additional check for null is not needed.

src/main/java/ognl/DefaultClassResolver.java
kazuki43zoo: I've changed to use generics.

src/main/java/ognl/DefaultClassResolver.java
kazuki43zoo: I refactored this code with your suggestion as reference.

kazuki43zoo:
@lukaszlenart Thanks for your review!
I've fixed your comments.
Please review again.

Thanks!

src/main/java/ognl/DefaultClassResolver.java
```diff
@@ -41,28 +42,27 @@
...
+            return result;
+        }
+        if (className.indexOf('.') == -1) {
+            String langClassName = "java.lang." + className;
+            result = Class.forName(langClassName);
+            classes.put(langClassName, result);
+        } else {
+            result = Class.forName(className);
+        }
+        classes.put(className, result);
```
lukaszlenart: This will duplicate `langClassName`, so we will end up with two entries for the same class.

kazuki43zoo: Thanks for your comment.
I think this behavior is same as the current version. Should change this behavior in this change?

lukaszlenart: Right, which means the `java.lang.` classes will be always located by the short version (`Integer` instead of `java.lang.Integer`). I think `.put` in line 61 can be thrown aways.

kazuki43zoo: I've fixed your comment. please review again.
Thanks.

src/main/java/ognl/DefaultClassResolver.java
```diff
@@ -41,28 +42,21 @@
...
-    	}
-	    return result;
-	}
+    public Class classForName(String className, Map context) throws ClassNotFoundException
+    {
+        Class result = classes.get(className);
+        if (result != null) {
+            return result;
+        }
+        result = (className.indexOf('.') == -1) ? Class.forName("java.lang." + className) : Class.forName(className);
```
lukaszlenart: I do prefer pure `if` clause but that's ok :)

lukaszlenart:
Do you want to port this fix into previous branches?

lukaszlenart:
LGTM 👍

kazuki43zoo:
Thanks for merging!

lukaszlenart:
3.2.5 is out and under way to the Central.

kazuki43zoo:
Wow, It's great!!

PR #49, opened 2018-03-20
Change return value to false form true on DefaultMemberAccess#setup method #47

### LINKED ISSUE #47: Is correct DefaultMemberAccess#setup?
Is correct following code?

https://github.com/jkuhnert/ognl/blob/0ca96030e40600892ea8b1548bbe275548ea1d48/src/test/java/ognl/DefaultMemberAccess.java#L114

Should be set the `Boolean.FALSE` in this case?
WDYT?

kazuki43zoo:
I've fixed #47 on 3.1.x branch.


lukaszlenart:
👍

lukaszlenart:
Let me know if are going to port another PR and I will wait with releasing 3.1.x

kazuki43zoo:
> if are going to port another PR

No need backport. You can release 3.1.x!

PR #50, opened 2018-03-20
Allow override a strategy for loading Class on DefaultClassResolver

kazuki43zoo:
I will propose to allow override a strategy for loading Class. Actually, the MyBatis core module want to override strategy using this extending point. Currently, we copy the `DefaultClassResolver` and modify a class loading logic.

WDYT?


lukaszlenart:
👍

lukaszlenart:
Should I release 3.2.6?

kazuki43zoo:
Yes! I wait to release 3.2.6 👍

lukaszlenart:
under way to the Central

PR #57, opened 2018-10-03
getReadMethod() should check default methods as well.

harawata:
This was fixed in master via #33, but was not backported to 3.1.x branch.


lukaszlenart:
LGTM 👍

lukaszlenart:
3.1.18 is under way to the Central :) Thanks a lot!

Issue #58, opened 2018-10-19
MemberAccess support private static method and private field, but do not support private static field.


hengyunabc:
MemberAccess support private static method and private field, but do not support private static field.

Demo:

```java
import java.lang.reflect.AccessibleObject;
import java.lang.reflect.Member;
import java.lang.reflect.Modifier;
import java.util.Map;

import ognl.ClassResolver;
import ognl.MemberAccess;
import ognl.Ognl;
import ognl.OgnlContext;
import ognl.OgnlException;

public class MyDemo {

	private static String statidField = "fff";

	private String privateField = "ppp";

	private static String staticFunc() {
		return "static func";
	}

	private static MyDemo newInstance() {
		return new MyDemo();
	}

	public static void main(String[] args) throws OgnlException {
		OgnlContext context = new OgnlContext(new ClassResolver() {
			@Override
			public Class classForName(String className, Map context) throws ClassNotFoundException {
				return Class.forName(className);
			}

		}, null, new DefaultMemberAccess(true));

		context.setTraceEvaluations(true);

		Object root = null;

		System.out.println(Ognl.getValue("@MyDemo@staticFunc()", context, root));
		System.out.println(Ognl.getValue("@MyDemo@newInstance().privateField", context, root));
		// exception
		System.out.println(Ognl.getValue("@MyDemo@statidField", context, root));
	}

	public static class DefaultMemberAccess implements MemberAccess {
		public boolean allowPrivateAccess = false;
		public boolean allowProtectedAccess = false;
		public boolean allowPackageProtectedAccess = false;

		/*
		 * ===================================================================
		 * Constructors
		 * ===================================================================
		 */
		public DefaultMemberAccess(boolean allowAllAccess) {
			this(allowAllAccess, allowAllAccess, allowAllAccess);
		}

		public DefaultMemberAccess(boolean allowPrivateAccess, boolean allowProtectedAccess,
				boolean allowPackageProtectedAccess) {
			super();
			this.allowPrivateAccess = allowPrivateAccess;
			this.allowProtectedAccess = allowProtectedAccess;
			this.allowPackageProtectedAccess = allowPackageProtectedAccess;
		}

		/*
		 * =================================================================== Public
		 * methods ===================================================================
		 */
		public boolean getAllowPrivateAccess() {
			return allowPrivateAccess;
		}

		public void setAllowPrivateAccess(boolean value) {
			allowPrivateAccess = value;
		}

		public boolean getAllowProtectedAccess() {
[...]

PR #59, opened 2018-10-19
MemberAccess support private static field. #58

hengyunabc:


src/java/ognl/OgnlRuntime.java
```diff
@@ -2076,11 +2076,26 @@ public static Object getStaticField(OgnlContext context, String className, Strin
...
+            Field f = getField(c, fieldName);
+            if (f == null) {
+                throw new NoSuchFieldException(fieldName);
+            }
             if (!Modifier.isStatic(f.getModifiers()))
                 throw new OgnlException("Field " + fieldName + " of class " + className + " is not static");
 
-            return f.get(null);
+            Object result = null;
+            if (!f.isAccessible() && context.getMemberAccess().isAccessible(context, null, f, null)) {
```
lukaszlenart:
```suggestion
            if (context.getMemberAccess().isAccessible(context, null, f, null)) {
```

The `!f.isAccessible()` is not needed IMO

src/java/ognl/OgnlRuntime.java
```diff
@@ -2076,11 +2076,26 @@ public static Object getStaticField(OgnlContext context, String className, Strin
...
+                context.getMemberAccess().setup(context, null, f, null);
```
lukaszlenart:
```suggestion
               Object state = context.getMemberAccess().setup(context, null, f, null);
```

Yous should store result of `setup()` and used it `restore()`

src/java/ognl/OgnlRuntime.java
```diff
@@ -2076,11 +2076,26 @@ public static Object getStaticField(OgnlContext context, String className, Strin
...
+                try {
+                    result = f.get(null);
+                } finally {
+                    context.getMemberAccess().restore(context, null, f, null, null);
```
lukaszlenart:
```suggestion
                    context.getMemberAccess().restore(context, null, f, null, state);
```

Restore the previous state

hengyunabc:
@lukaszlenart Thanks, I updated the commit.

src/java/ognl/OgnlRuntime.java
```diff
@@ -2076,11 +2076,26 @@ public static Object getStaticField(OgnlContext context, String className, Strin
...
+            Object result = null;
+            if (context.getMemberAccess().isAccessible(context, null, f, null)) {
+                Object state = context.getMemberAccess().setup(context, null, f, null);
+                try {
+                    result = f.get(null);
+                } finally {
+                    context.getMemberAccess().restore(context, null, f, null, state);
+                }
+            } else {
+                result = f.get(null);
```
lukaszlenart: Now I wonder if we shouldn't throw an exception here, I mean, the field is not accessible but we want to get access to it anyway.

hengyunabc: @lukaszlenart  Do you mean line 2079? 

```java
            Field f = getField(c, fieldName);
            if (f == null) {
                throw new NoSuchFieldException(fieldName);
            }
```

`OgnlRuntime#getField` return all fields, contain private fields, so throw `NoSuchFieldException` is ok.

lukaszlenart: No, I meant that we check access first and if the field is not accessible we access it anyway. This can lead to a security vulnerability as basically we ignore if the field is accessible or not.

hengyunabc: @lukaszlenart I got it, I updated the commit.

lukaszlenart:
LGTM 👍

lukaszlenart:
Thanks a lot!

lukaszlenart:
Do you plan to cherry-pick those changes into the `master` branch?

hengyunabc:
> Do you plan to cherry-pick those changes into the `master` branch?

OK, I created a new PR: #60. 

In addition, may I ask is there any plan to release the new version?

lukaszlenart:
In few hours, working on it ;-)

PR #67, opened 2019-01-21
Fix backwards-compatibility issue introduced with PR#59/#60

JCgH4164838Gh792C124B5:
Fix backwards-compatibility issue introduced with PR#59/#60
 - Made access to non-public static fields _an optional opt-in feature_ (control flag via `context`).
 - Restored _original behaviour when accessing public static fields_ (the default behaviour).
 - Updated unit tests to test proper behaviour with both original (default) and opt-in behaviour.

### DIFF
--- a/src/java/ognl/ASTStaticField.java
+++ b/src/java/ognl/ASTStaticField.java
@@ -89,15 +89,21 @@ public class ASTStaticField extends SimpleNode implements NodeType
                 result = true;
             } else
             {
-                Field f = OgnlRuntime.getField(c, fieldName);
-                if (f == null) {
-                    throw new NoSuchFieldException(fieldName);
+                Field f;
+                try {
+                    f = c.getField(fieldName);  // Public fields checked first (direct)
+                } catch (NoSuchFieldException nsfe) {
+                    f = OgnlRuntime.getField(c, fieldName);  // Non-public fields checked (access controlled)
+                    if (f == null) {
+                        throw new NoSuchFieldException(fieldName);
+                    }
                 }
+                final int fModifiers = f.getModifiers();
 
-                if (!Modifier.isStatic(f.getModifiers()))
+                if (!Modifier.isStatic(fModifiers))
                     throw new OgnlException("Field " + fieldName + " of class " + className + " is not static");
 
-                result = Modifier.isFinal(f.getModifiers());
+                result = Modifier.isFinal(fModifiers);
             }
         } catch (ClassNotFoundException e) {
             reason = e;
--- a/src/java/ognl/OgnlRuntime.java
+++ b/src/java/ognl/OgnlRuntime.java
@@ -2099,24 +2099,34 @@ public class OgnlRuntime {
                 }
             }
 
-            final Field f = getField(c, fieldName);
-            if (f == null) {
-                throw new NoSuchFieldException(fieldName);
+            Field f;
+            try {
+                f = c.getField(fieldName);  // Public fields checked first (direct)
+            } catch (NoSuchFieldException nsfe) {
+                f = OgnlRuntime.getField(c, fieldName);  // Non-public fields checked (access controlled)
+                if (f == null) {
+                    throw new NoSuchFieldException(fieldName);
+                }
             }
-            if (!Modifier.isStatic(f.getModifiers())) {
+            final int fModifiers = f.getModifiers();
+            if (!Modifier.isStatic(fModifiers)) {
                 throw new OgnlException("Field " + fieldName + " of class " + className + " is not static");
             }
+            final Object result;
 
-            Object result = null;
-            if (context.getMemberAccess().isAccessible(context, null, f, null)) {
-                final Object state = context.getMemberAccess().setup(context, null, f, null);
-                try {
-                    result = f.get(null);
-                } finally {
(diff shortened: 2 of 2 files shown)


lukaszlenart:
I'm not sure if this a good solution, OGNL should have such control flags and as far I recall this will be the first flag ever.

JCgH4164838Gh792C124B5:
Hello @lukaszlenart .

If using a control flag for the behaviour is not desirable, what do you think about a "partial return" to the original behaviour when dealing with `public static` fields ?  
Basically attempt `c.getField()` first (which can only succeed for `public static` fields), then if that fails attempt @hengyunabc 's methodology for non-public fields (which performs the access checks).

Such a scheme would have a couple of advantages:
- avoids a control flag being introduced.
- retains backwards compatibility (3.1.18 and earlier) for `public static` fields (which should be "safe" to access values for as they are declared public anyway).  Should allow for better "drop-in" compatibility with existing applications.
- retains access control for non-public static fields.

Tradeoffs:
- slightly more efficient access for `public static` fields (probably the majority of use cases).
- slight more expensive access for non-public static fields (initial public-only check will fail).

I could _amend this commit to match the above_ so that you can take a look.

Would you like this PR to be an example of that strategy instead ?

JCgH4164838Gh792C124B5:
Hello.
It didn't take too long to remove the control flag and implement the alternate backward-compatible strategy described above.  The previous commit was amended to achieve this.
Let me know if it seems better-suited.

lukaszlenart:
Nice work, looks far better 👍 
@hengyunabc & @aleksandr-m what do you think? Looks good to me.

src/java/ognl/ASTStaticField.java
```diff
@@ -89,15 +89,23 @@ public boolean isNodeConstant(OgnlContext context)
...
+                    f = c.getField(fieldName);  // Public fields checked first (direct)
+                } catch (NoSuchFieldException nsfe) {
+                    f = OgnlRuntime.getField(c, fieldName);  // Non-public fields checked (access controlled)
+                    if (f == null) {
+                        throw new NoSuchFieldException(fieldName);
+                    }
                 }
+                fModifiers = f.getModifiers();
 
-                if (!Modifier.isStatic(f.getModifiers()))
```
aleksandr-m: Why change this?

JCgH4164838Gh792C124B5: Hello @aleksandr-m .
It was just a minor optimization (equivalent to one already used in `OgnlRuntime`) to replace _two calls_ to `f.getModifiers()`:
.. if (!Modifier.isStatic(f.getModifiers())) / result = Modifier.isFinal(f.getModifiers());
with a _single call/assignment_ for the field modifiers `fModifiers = f.getModifiers()`: 
.. if (!Modifier.isStatic(fModifiers)) / result = Modifier.isFinal(fModifiers).

aleksandr-m: `getModifiers()` is just a getter it is ok to use it directly. If you want to keep `fModifiers` then declaration and assignment can be moved to them same line, currently it looks very odd.

JCgH4164838Gh792C124B5: Hello @aleksandr-m .
Thanks for the feed-back and suggestion.
The declaration/assignment for `fModifiers` has been moved to the same line in both modules (and the declaration of one variable moved closer to its usage point).  Hopefully it looks a little cleaner now ?

aleksandr-m:
Looks good. :+1:

lukaszlenart:
Ok, no objections so let's merge and release a new version :)

lukaszlenart:
What do you think about putting this back in OGNL 3.2.x (checking access to static fields) and add an additional flag to Struts 2.6 to enable/disable such checking in `SecureMemberAccess`?

JCgH4164838Gh792C124B5:
Hello @lukaszlenart .

I guess the question for OGNL 3.2.x is whether it is preferable to:
1) Always allow access to public static fields (as with 3.1.x)  ... or ...
2) Always perform access checks (for both public and non-public static fields).

Choice 1) would be more consistent (as 3.2.7/earlier also provided that behaviour), but choice 2) would provide more "fine-grained" control over access.  I'll put together a PR matching choice 1 for now and it can be accepted/rejected depending on what direction you (and others) think it should go.

An additional flag or flag(s) in Struts 2.6 to enable/disable checks in `SecureMemberAccess` might be a good feature, but I wonder if for it to be effective it might need to be backed with a separate inclusion/exclusion set just for the fields (since method/field access might not line up 1-to-1) ?

JCgH4164838Gh792C124B5:
Hello @lukaszlenart .

OK, sounds good.  In that case doesn't that mean that the current access check flows for the OGNL 3.2.x code should remain in place (as with 3.2.8+), if I'm understanding correctly ? 
 
If so, then maybe the only things that could change would be updates to ASTStaticField and OgnlRuntime that restore the type-of-exception thrown if the field is non-static (throw `OgnlException` instead of `NoSuchFieldException`).  That might be unnecessary if the behaviour for access to public static fields has changed anyway...  
Also to provide the original exception type would still require retrieving the field twice (once directly, to get modifiers, then if static, again to enforce access checks) ... unless I'm missing something ?

PR #69, opened 2019-03-18
invoke method in a fine grained java sandbox

yasserzamani:
Applies a customized java SecurityManager on method invokes which disables ognl evaluations to do most top dangerous actions (execute,exit,thread access,... - see OgnlSecurityManager's overloaded methods). However, also honors previous security manager and policies if any set, as parent, and rolls back to them after ognl evaluation finished.

P.S. Struts 2.5.21-SNAPSHOT passes all tests with this ognl 💯 😃


src/java/ognl/OgnlRuntime.java
```diff
@@ -899,12 +913,53 @@ public static Object invokeMethod(Object target, Method method, Object[] argsArr
...
 
-            result = method.invoke(target, argsArray);
+            result = invokeMethodInJavaSandbox(target, method, argsArray);
         }
 
         return result;
     }
 
+    private static Object invokeMethodInJavaSandbox(Object target, Method method, Object[] argsArray) throws InvocationTargetException, IllegalAccessException {
+        SecurityManager parentSecurityManager = System.getSecurityManager();
```
lukaszlenart: Why not to use `this. _securityManager`? Someone could already re-define the SecurityManager used by OGNL (OGNL already supports checking permissions)

src/java/ognl/OgnlRuntime.java
```diff
@@ -899,12 +913,53 @@ public static Object invokeMethod(Object target, Method method, Object[] argsArr
...
             }
 
-            result = method.invoke(target, argsArray);
+            result = invokeMethodInJavaSandbox(target, method, argsArray);
         }
 
         return result;
     }
 
+    private static Object invokeMethodInJavaSandbox(Object target, Method method, Object[] argsArray) throws InvocationTargetException, IllegalAccessException {
```
lukaszlenart: I would name this method as `invokeMethodInSandbox`

src/java/ognl/OgnlRuntime.java
```diff
@@ -899,12 +913,53 @@ public static Object invokeMethod(Object target, Method method, Object[] argsArr
...
+        try
+        {
+            if (parentSecurityManager != null) {
+                parentSecurityManager.checkPermission(new SecurityPermission("getPolicy"));
+                parentSecurityManager.checkPermission(new SecurityPermission("setPolicy"));
+                parentSecurityManager.checkPermission(new RuntimePermission("setSecurityManager"));
+            }
+            parentPolicy = Policy.getPolicy();
+        } catch (SecurityException ex) {
+            return method.invoke(target, argsArray);
```
lukaszlenart: Not sure if I understand this code, you will invoke a method anyway if there was a `SecurityException`? I would assume it's not allowed to invoke the method, though.

src/java/ognl/OgnlRuntime.java
```diff
@@ -899,12 +913,53 @@ public static Object invokeMethod(Object target, Method method, Object[] argsArr
...
+
+            try {
+                return method.invoke(target, argsArray);
+            } catch (SecurityException ex) {
+                ex.printStackTrace();
+                throw new IllegalAccessException("Method [" + method + "] cannot be accessed.");
+            } finally {
+                System.setSecurityManager(parentSecurityManager);
+                Policy.setPolicy(parentPolicy);
+            }
```
lukaszlenart: I have no idea what's going on here :(

src/java/ognl/OgnlRuntime.java
yasserzamani: Yes I saw, but with `this. _securityManager` it seems user only can say OGNL if invocation of e.g. `some.package.SomeClass.someMethod` is permitted or not. OK it's nice and I kept it.

But my security manager *then in a higher level* guarantees that any invocation *doesn't reach* top most dangerous actions (e.g. execute,exit,thread access and etc) even if e.g. an inner of inner of inner method is going to do that action. It's an awesome JDK security model that guarantees this! For example if an attacker was successful to inject `someField=new some.third.party.package.that.target.app.uses.SomeClass().getScriptEngine().run('cat pwd')` then my security manager doesn't allow this (even if user has forgot to add this to blacklist!) because JDK security model detects that it finally reaches an execution and throws an exception for me 😃 

At bottom, despite it seems a massive change, I think it's fine and OK to add it to all branches because firstly it's really a simple but great security improvement IMHO and secondly if it breaks something in user apps, then they will be happy to discover because that thing is a security flaw - my security manager only blocks top most dangerous actions (see overloaded methods) and relaxes with others and user's apps must already satisfy this rule.

src/java/ognl/OgnlRuntime.java
yasserzamani: As I use and honor [JDK's internal default sandbox model](https://docs.oracle.com/javase/7/docs/technotes/guides/.../security-spec.doc1.html) I named such that to refer reader to JDK sandbox (not our own or third-party sandbox impel).

src/java/ognl/OgnlRuntime.java
yasserzamani: I saw in my tests when `Policy.getPolicy()` throws `SecurityException` it means there is already a policy (e.g. my policy applied in previous recursions) that doesn't allow get policy. So we have to honor user higher applied policy and just invoke the method which occurs in and honor higher policy context.

src/java/ognl/OgnlRuntime.java
yasserzamani: When OGNL's invocation finished with our policy and security manager, then here restore all things to what was before method invocation because here we just finished our sensitive action so we relaxes JVM - security manager and policy are global JVM and affect all threads and processes of current JVM!

harawata:
Thanks for the heads up, @lukaszlenart !
And thank you @yasserzamani for your contribution to make Struts more secure!

While this change may improve Struts security, OGNL is used in various libraries/frameworks and it might not be ideal to some of them.
MyBatis, for example, does not need to worry about users' input being evaluated by OGNL and using SecurityManager may affect performance.

Have you considered making this change in Struts instead?

yasserzamani:
@harawata thanks a lot for the feedback! yes I was aware and expected such requests but I didn't have enough insight and estimate about how many references OGNL does have (except Struts)? and if my security manager is OK for them also? So I decided to try here firstly because IMHO usually no body expect OGNL to execute another executable or access other threads and etc (I just only check such things and relax on other less important things - see overloaded methods).

Regarding performance, I had no time due an incoming travel but I was going to cache methods who were successful in their first invoke (i.e. no `SecurityException` catch for them). Then I don't set security manager and policy for them in next calls.

Anyway if it didn't accepted here then yes, I think it should be possible to move to Struts. I didn't check but I feel logically it should be but I think it cannot be as nice as here because in Struts we cannot wrap the `invokeMethod` only and will have to change all reference to OGNL runtime :(

yasserzamani:
Oh, caching isn't safe always. Maybe a method doesn't reach `SecurityException` with some input or environment parameters but does reach with other.

Now I have a better idea. I'm going to add a new functionality which allows user to enable sand-boxing. I put it disabled by default. Then Struts can enable it.

I also am going to make sand-boxing (security manager and policy) implementations inject-able which enables users to inject their own implementation instead of my default implementation.

JCgH4164838Gh792C124B5:
Hello @yasserzamani and @lukaszlenart .

I haven't had time to look at the PR in detail, so my thoughts and opinions below might change after a more thorough review.  :)

Kudos to @yasserzamani for looking into a sandboxing mechanism for OGNL.  :)

@harawata raises some points worth considering, and the _potential for a negative performance impact being one of them_.  The synchronization on `Policy.class` _looks like it could potentially cause major thread-contention_ (and the attempts to install-invoke-restore `SecurityManager` instances seems it might be risky ...).

It's an interesting sandboxing idea for sure, but if it goes ahead it should probably be an _"opt-in"_ feature that OGNL library users can decided whether to enable or not (as @yasserzamani mentioned in his last comment).  If the feature is _"opted-out"_ (disabled), hopefully it could have little-to-no overhead when running "disabled".  OGNL Library users would also want to be aware that the library may install a SecurityManager too.

Even without the changes offered in this PR, developers have the option to implement their own Java SecurityManager implementation to permit or restrict operations (including OGNL and any other libraries running in the application/JVM context).  A framework (like Struts) could check if a `SecurityManager` is present, and if not perform a log warning on initialization recommending that the developer/user consider installing one for improved security...

JCgH4164838Gh792C124B5:
Hello again @lukaszlenart  and @yasserzamani .

The type of change(s) proposed for this PR might be too risky for 3.1.x, unless an _opt-out exists and provides equivalent performance_ to previous 3.1.x versions.

After thinking about it some more it feels like a `SecurityManager`-driven sandboxing might be too low-level to apply to OGNL itself (seems like a _higher-level concern_ for an application/container to handle, and the _performance concerns remain_).

OGNL is already performing method-permission checks if a `SecurityManager` is installed, maybe that should be the limit for scope and performance reasons.

Instead of looking at sandboxing within OGNL itself, maybe an _opt-in_ / _opt-out_ "expression guard" mechanism for `Ognl` itself which checks for "suspicious" expressions and rejects them when they are detected.  The (_optional_) "expression guard" check could be performed as the 1st operation 
in `Ognl.parseExpression`, [here](https://github.com/jkuhnert/ognl/blob/82889e1a07820b3df32efe4c90f010fea278a34b/src/java/ognl/Ognl.java#L108-L111) before an `OgnlParser` is even constructed.  Performance implications would have to be considered in such a solution too.  A custom (application-provided) decision-check method might be an option too.  

Maybe something like that would be of use or a potential alternative idea ?  The concerns for the 3.1.x line would be the same for it as well ...

JCgH4164838Gh792C124B5:
Hello @yasserzamani .
Inspired by your PR, I attempted to come up with an "expression guard" mechanism, but could not think of any implementation that might be remotely efficient (too many possible combinations for expressions).  
An alternative idea was put in PR#71 to provide an opt-in method deny list (using Method matching similar to the existing cache mechanism).  Feel free to take a look and let me know what you think (its implementation raises the same potential performance concerns as a sandboxing mechanism does).

harawata:
Hi all,

I took a quick look at Struts' source and there seemed to be no direct call to `OgnlRuntime#invokeMethod`, so I assumed it was called via `OgnlRuntime#callAppropriateMethod` or `OgnlRuntime#getMethodValue`.
And these methods check the result of `MemberAccess#isAccessible` before calling `invokeMethod`.

It makes me wonder if it is cleaner to implement the blacklist filter in Struts' [MemberAccess implementation](https://github.com/apache/struts/blob/master/core/src/main/java/com/opensymphony/xwork2/ognl/SecurityMemberAccess.java) as there is overlap in these concepts.

yasserzamani:
Hi all,


According to previous comments, it seems I should explain the philosophy
behind this PR which is an answer to previous comments.


This PR tries to secure OGNL users against "expression injection" attack
onetime forever. By onetime forever, I mean e.g. Struts already has an
internal sandbox mechanism via exclusion/deny lists/expressions, but the
time shows it doesn't work well. The reason is, say X is an OGNL user
(has OGNL in it's dependencies) and say X has other M, N, O, P, Q, R, S
and T dependencies. In Struts I learnt when X has an unknown "expression
injection" security flaw, it's difficult and somehow impossible to
discover and add all possible attack vectors from M, N, O, P, Q, R, S
and T to exclusion/deny lists/expressions. It makes it harder and
impossible maintaining them over time where M, N, O, P, Q, R, S and T
change over time.


So "expression guard" and "deny list" aren't needed and aren't an
alternative for this PR. If they were enough and were working then I
didn't open this PR as we already have them in Struts. They aren't
enough and don't work because as I said in previous paragraph, it's
impossible to discover and maintain all possible attack vectors over time.


About asking end user to implement and apply it's own SecutiyManager and
Policy, it doesn't look good for me because has more bigger performance
decrease -- I'm going to make it optional and disabled by default. mine
is not active always. If enabled, it's active only during method
invocation by OGNL which are not expected to be complex methods with
complex and sensitive activities. But end user implementation has to be
active always during whole application run. Additionally I think it will
need a lot of efforts to decide if current stacktrace e.g. has execution
permit or not because it's an application wide applied manager.


P.S. Actually OGNL user e.g. X should fix and shouldn't have any
"expression injection" flaw but this PR tries to provide an option to X
to enable a security manager when X afraid if it can has an unknown
"expression injection" flaw and afraid it cannot has a comprehensive
exclusion/deny lists/expressions. Then this PR disables any sensitive
action e.g. execution even if attacker had a successful "expression
injection" on X (OGNL user) in any unknown way.

JCgH4164838Gh792C124B5:
Thanks for the detailed explanation and reasoning for the approach in this PR.  :)
The idea is neat/cool, but the performance impact concern remains.  :)

Putting aside overhead from setting/restoring `SecurityManager` and `Policy` objects on a "per-call basis", a concern remains that `invokeMethodInJavaSandbox()` _appears to always limit OGNL concurrency to **only one (1) thread**_ (due to synchronization on the `Policy` class).
The old `invokeMethod()` seems to only limit concurrency to "_at worst_ 1 thread per `Method` type".

yasserzamani:
Yes you're right. I think we can check and skip sync, reset and restore
where there are already an applied OGNL security manager and policy exists.

yasserzamani:
Hi all,

It seems I managed to address concerns and issues mentioned above by above commit (make optional).

I think now we (me or anyone) can test performance via JMeter and compare results (response time, fails e.g. concurrent issues if any and errors e.g. timeouts) between disabled and enabled sandbox on high concurrent many threads and a long valid expression e.g. "some.thing.some.other.thing.someMethod()...".

Thanks in advance!

yasserzamani:
Hi all,

I had missed some points and James was absolutely right, my tests showed that concurrency performance was too bad with my previous implementation!

But now I also addressed them with above commit (fix and add test for concurrency performance and issues). I think now I can conclude my work with following report:

* It's optional and disable by default so it's backward compatible with no performance decrease.
* Added unit-test shows response time of all 1000 concurrent threads with random 0-3 seconds method body weight is about 3 seconds and all successfully blocked i.e. performance is OK and also an attacker cannot bypass sandbox with high concurrent requests :)
* Struts 2.5 tests which have a lot of OGNL calls pass with following total times:
Disabled Sandbox (OGNL Default): [INFO] Total time:  03:08 min
Enabled Sandbox: [INFO] Total time:  03:12 min
Which means it looks good to me in a serial manner also.

Kind Regards.

JCgH4164838Gh792C124B5:
Hello @yasserzamani .
Thanks for the changes and effort you have put towards this PR. 
It will take some time to "absorb" the changes, try/test things out and give some feedback.  I'll add this to my "stack" and try to get back to you (others may be able to provide feedback sooner).
Good work.  :smile:

yasserzamani:
Thanks a lot for your support! Last night while I re-thought I felt potential possible concurrency issue between install and uninstall sandbox. Now I fixed that also, plus some improvements :) - performance still looks good and is same as previous report.

yasserzamani:
Hi all,

With last 2 commits, Struts 2.5 tests (enabled sandbox) pass except 5 tests below. 5 of 1828 test i.e. 0.2% means they should be fixed at Struts side I think so above new two commits looked GTM at OGNL side.

Regards.

>[ERROR] Failures: 
>[ERROR]   AnnotationXWorkConverterTest.testFindConversionErrorMessage:173 expected:<Invalid [date for birth].> but was:<Invalid [field value for field "birth"].>
>[ERROR]   XWorkConverterTest.testFindConversionErrorMessage:213 expected:<Invalid [date for birth].> but was:<Invalid [field value for field "birth"].>
>[ERROR] Errors: 
>[ERROR]   ActionsFromSpringTest.testChainingProxiedActions:105 » AccessControl access de...
>[ERROR]   ActionsFromSpringTest.testProxiedActionIsNotAccessible:137 » AccessControl acc...
>[ERROR]   OptGroupTest.testOptGroupWithValueKey:249 » Struts
>[INFO] 
>[ERROR] Tests run: 1823, Failures: 2, Errors: 3, Skipped: 0

src/java/ognl/OgnlRuntime.java
```diff
@@ -882,7 +883,7 @@ public static Object invokeMethod(Object target, Method method, Object[] argsArr
```
JCgH4164838Gh792C124B5: Current 3.1.23 changes could be incorporated here.

src/java/ognl/security/MethodBodyExecutionSandbox.java
```diff
@@ -0,0 +1,158 @@
...
+import java.security.Policy;
+import java.security.SecurityPermission;
+
+/**
+ * Guarantee a singleton shared thread-safe security manager and sandbox for user's methods body execution
+ * @author Yasser Zamani
+ * @since 3.1.23
+ */
+public class MethodBodyExecutionSandbox {
+    private static Permissions userDemandPermissions;
```
JCgH4164838Gh792C124B5: I think that these flag and permission variables either need to be made volatile, or all the access points to them may need to by synchronized.

src/java/ognl/security/MethodBodyExecutionSandbox.java
```diff
@@ -0,0 +1,158 @@
...
+import java.security.Permissions;
+import java.security.Policy;
+import java.security.SecurityPermission;
+
+/**
+ * Guarantee a singleton shared thread-safe security manager and sandbox for user's methods body execution
+ * @author Yasser Zamani
+ * @since 3.1.23
+ */
+public class MethodBodyExecutionSandbox {
```
JCgH4164838Gh792C124B5: I think that these flag and permission variables either need to be made volatile, or all the access points to them may need to by synchronized.

src/java/ognl/security/MethodBodyExecutionSandbox.java
```diff
@@ -0,0 +1,158 @@
...
+    private static Permissions userDemandPermissions;
+    private static Policy userDemandPolicy;
+    private static SecurityManager userDemandSecurityManager;
+    private static boolean enabled;
+    private static boolean disabled;
+    private static Integer residentsCount = 0;
```
JCgH4164838Gh792C124B5: It looks like no code needs residentsCount to be an "Integer" (object), might be worth making it a primitive "int" and avoid autoboxing overhead.

src/java/ognl/security/OgnlPolicy.java
```diff
@@ -0,0 +1,56 @@
+package ognl.security;
+
+import java.lang.reflect.ReflectPermission;
```
JCgH4164838Gh792C124B5: This may be an unused import (IDE reported)

src/java/ognl/security/OgnlPolicy.java
```diff
@@ -0,0 +1,56 @@
...
+import java.security.*;
+import java.util.Enumeration;
+
+/**
+ * Wraps current policy with minimum permissions plus user demand further permissions
+ * @author Yasser Zamani
+ * @since 3.1.23
+ */
+public class OgnlPolicy extends Policy {
```
JCgH4164838Gh792C124B5: These two fields could be made final (IDE suggested).

src/java/ognl/security/OgnlSecurityManager.java
```diff
@@ -0,0 +1,333 @@
...
+import java.lang.reflect.ReflectPermission;
+import java.net.InetAddress;
+import java.security.Permission;
+
+/**
+ * Wraps current security manager with sensitive actions e.g. exit and exec disabled if is inside OgnlRuntime
+ * @author Yasser Zamani
+ * @since 3.1.23
+ */
+public class OgnlSecurityManager extends SecurityManager {
```
JCgH4164838Gh792C124B5: This field could be made final (IDE suggested).

src/test/java/ognl/TestOgnlRuntime.java
```diff
@@ -172,6 +178,111 @@ public void test_Call_Method_VarArgs()
...
+
+    public void test_Call_Method_In_JDK_Sandbox_Thread_Safety()
+            throws Exception {
+        final OgnlContext context = (OgnlContext) Ognl.createDefaultContext(null);
+        final GenericService service = new GenericServiceImpl();
+
+        MethodBodyExecutionSandbox.enable(null, null, null);
+
+        try {
+            final int NUM_THREADS = 1000;
```
JCgH4164838Gh792C124B5: **Note of interest**:  It appears this PR's OGNL sandbox may reveal some kind of thread limitation that shows up under JDK 7 (1.7.0_79 32-bit),  The 32-bit JVM threw failures in the test due to inability to create native threads, but it did not happen when not using the sandbox.  The test could complete after dropping the NUM_THREADS below 350 (approximately) under the 32-bit JVM with the sandbox enabled.
Switching to a 64-bit JVM (1.7.0_79 64-bit) made the native thread failures disappear when running under the sandbox.  It could be simply a factor of processing speed or differences in the 32/64-bit JVM implementations ...

lukaszlenart: Nice catch ... I wonder how many users are still based on 32-bit JVM, there is rather a limited number though.

yasserzamani:
Hi again,

I addressed issue raised recently above :)

During my further thinks and research, I realized I have to deny all
permissions except package access because each of them can be used for
mortal attacks! I found a scenario for each one.

Now 3% of Struts tests fail but I'm happy to catch them! Each one can be
used for mortal attacks if attacker find a way for expression injection!

So this PR is **still under work and thinking** - to find a way to
enable user to use something like JDK's `doPrivilaged` or anything that
can minimize possible backward-compatibility while keeping same security.

Regards.

yasserzamani:
Hi @JCgH4164838Gh792C124B5 , thanks for your interest and review!

> You might consider incorporating recent changes to OgnlRuntime.java from OGNL 3.1.23 into this PR's version (in the event the PR is accepted this OgnlRuntime.java will overwrite those changes).

I didn't understand, I'm currently synced with latest 3.1 branch. If you meant your other PR, it's not a good practice to copy your changes here. The second accepted PR will merge and sync itself with first accepted one once needed.

> Have you determined why the current implementation (circa April 13th) for this PR is now failing with the regression of Struts 2.5.21-SNAPSHOT when the sandboxing is enabled ? The earlier implementation (circa April 7th) still passed that regression without any failures/errors.

Yes it's because I dropped `supressAccessCheck` and `getProtectionDomain` permissions because I realized they are also dangerous. For example `supressAccessCheck` can be used to set enabled=false.

> I tried stepping through some of the failures and it seems the current PR's sandboxing is blocking constructor access for some calls where it wasn't before ?

Yes I personally didn't expect these fails. I thought Struts uses OGNL only for apply http parameters on actions and vice versa, action to jsp tags. But I realized it does more complex things with OGNL e.g. `getText` which asks a TextProvider from container then container needs `supressAccessCheck` to inject dependencies then fails. For another example Struts even calls user actions via OGNL.

Recently I blocked all permissions and about 3% (40) tests fail. Putting a break-point on `MethodBodyExecutionSandbox.accessDeny` helps a lot finding out why each one fails.

> Concern: Since the current sandboxing implementation fails (when enabled) the existing Struts 2.5.21-SNAPSHOT regression suite, does that present a risk of introducing unexpected behaviour to Struts if a new OGNL version using this mechanism was introduced ?

Yes but it depends **how much cleanly** we are able to handle these and adapt Struts to at least pass all tests **cleanly**.

> With that consideration, do you think it might be less risky to consider this change for OGNL 3.2.x and Struts 2.6.x (where "breaking behaviour" would be more expected) ?

I think it depends **how much cleanly, robustly, standard-like and JDK-like** we are able to implement new OGNL and adapt new Struts. Say Struts 2.5 user reports a break. If that break was because of our weakness in sandbox implementation or Struts adaption then yes it's too bad. But if not, the user will be so happy and surprised because this break reveals a security flaw in underlying app and will fix it which makes app more secure.

Anyway, while I think we finally will be able to deliver clean, robust, standard-like and JDK-like sandbox, then the rest success is more dependent to how much complex things Struts does via OGNL and how much we are able to adapt them with new OGNL - @lukaszlenart despite it's complexity, I think it's not a waste work because it makes Struts more secure to not use OGNL for complex things and so, will preserve Struts against future unknown expression injection flaws.

src/test/java/ognl/TestOgnlRuntime.java
yasserzamani: :+1: fixed via scaling down by 10.

src/java/ognl/security/OgnlSecurityManager.java
yasserzamani: :+1: fixed

src/java/ognl/OgnlRuntime.java
yasserzamani: :confused: but I'm synced with latest 3.1 branch!

src/java/ognl/security/MethodBodyExecutionSandbox.java
yasserzamani: deleted due to not being meaningful enough. Will be added in future if user demand.

src/java/ognl/security/MethodBodyExecutionSandbox.java
yasserzamani: :+1: Firstly I used it for sync but next I used .class but forgot to revert this. thanks! :)

src/java/ognl/security/OgnlPolicy.java
yasserzamani: What's your IDE? It wasn't unused. I had `...instance of...` for that.

src/java/ognl/security/OgnlPolicy.java
yasserzamani: Deleted due to not being meaningful enough. Will be added in future if user demand.

src/java/ognl/security/OgnlPolicy.java
JCgH4164838Gh792C124B5: I'm using NetBeans 8 currently.  The import does appear to be unused as `ReflectPermission` only appears in the import statement itself ?

Commenting the import out, it still compiled for me (with my slightly-out-of-date local version).  Looks like `ReflectPermission` only appears in the import in the latest revision too.

src/java/ognl/OgnlRuntime.java
JCgH4164838Gh792C124B5: Please ignore the comments about the "upper-block" differences (I was getting confused with some local changes in many back-and-forth copying) :blush:

For `invokeMethod`  changes in 3.1.23 vs. this PR, I think Git sees the "whole block" changing (so no conflict per se).  Performing a manual compare it looks like the new logic in this PR's invokeMethod uses the original 3.1.22 logic (so it would overwrite the usage of `methodAccessCacheValue`, `methodPermCacheValue` in 3.1.23).  At least I think so ? :smiley:

I _just noticed_ that the `@since` blocks in the PR are at `3.1.23`, so they could be changed to `3.1.24` too ?

yasserzamani: Oh sorry you were right. There is a recent release. Thanks!

src/java/ognl/security/OgnlPolicy.java
yasserzamani: Oh sorry you were right. I confused it with OgnlSecurityManager file :(

JCgH4164838Gh792C124B5:
This PR certainly highlights a case of _balancing the power of the OGNL expression language and some security implications_ of that power.

There are (sometimes) _legitimate use cases_ for various complex expressions (e.g. calling `getText()`), especially for local (non-public) or enterprise OGNL usage in an application/framework.  For that reason software/frameworks using OGNL (e.g. Struts) may want to provide _opt-in/opt-out mechanisms at various scopes_ (e.g. global, tag, action, etc.) in order to use "classic OGNL" for backwards-compatibility.

There may still be some value in permitting users the _flexibility to decide some usage, based on the individual application needs_ (and risk-tolerance).

src/java/ognl/security/MethodBodyExecutionSandbox.java
```diff
@@ -0,0 +1,87 @@
...
+ * execution finished.</p>
+ *
+ * @author Yasser Zamani
+ * @since 3.1.23
+ */
+public class MethodBodyExecutionSandbox {
+    private static int residentsCount = 0;
+
+    private static SecurityManager parentSecurityManager;
+
```
JCgH4164838Gh792C124B5: To minimize overhead with `System.getProperty("ognl.security.manager")` calls and to avoid possible SecurityException at unexpected times, maybe consider setting a one-time-flag using a static initializer ?  Something like:
```java
    private static final boolean sandboxEnabled;
    static {
        boolean tempSandboxEnabled = false;
        try {
            tempSandboxEnabled = (System.getProperty("ognl.security.manager") != null);
        } catch (Exception ex) {
            // Ignore property read failure
        } finally {
            sandboxEnabled = tempSandboxEnabled;
        }
    }
``` 
and refer to that sandboxEnabled flag instead.  Unfortunately it makes unit testing more complicated.

src/java/ognl/security/MethodBodyExecutionSandbox.java
```diff
@@ -0,0 +1,87 @@
...
+        }
+
+        try {
+            return method.invoke(target, argsArray);
+        } finally {
+            leave();
+        }
+    }
+
+    private static boolean enter() {
```
JCgH4164838Gh792C124B5: Suggestion:  Move `getProperty()` check outside of synchronization block for better performance:
```java
                if (System.getProperty("ognl.security.manager") == null) {
                    // isn't enabled so simply just invoke the method outside sandbox
                    return false;
                }
                synchronized (MethodBodyExecutionSandbox.class) {
                    if (installSandboxIntoJVM()) {
                // Rest of code
```
If the `sandboxEnabled` flag idea is used, replace the `if (getProperty())` check with `if (!sandboxEnabled)`.

yasserzamani: It must be inside `if (residentsCount == 0) {` elsewhere it throws exception - already our own installed security manager doesn't allow reading any property :) -- `residentsCount == 0` means there is no already our own installed security manager.

src/java/ognl/security/MethodBodyExecutionSandbox.java
yasserzamani: I would like to keep this check lazy to allow user enable/disable it programmatically but it's a good idea to surround `getProperty` with try/catch to honor upper level security manager, I'll 👍

yasserzamani:
wow! now this is a very small and optimized code and still same result on Struts (about 2% i.e. 40 tests fails only on core) - however with `-Dsun.reflect.inflationThreshold=2147483647` JVM arg which currently is required :(

Via local logging I realized that this will be decreased to only 7 fails if I was able to handle file read permission, and this will be eliminated if I also was able to adapt Struts `getText...container injections` after then. I feel I will :)

src/java/ognl/security/MethodBodyExecutionSandbox.java
JCgH4164838Gh792C124B5: Hello @yasserzamani .

If I'm reading the code correctly, if `System.getProperty("ognl.security.manager")` has to be inside `synchronized (MethodBodyExecutionSandbox.class) ` it's _going to force a thread-synchronization before every OGNL method invocation_ ... _even if the library user doesn't want the sandbox enabled_ (system property false).  :(

From what I can see that creates an issue of overhead/thread synchronization that the library user has to pay even if they have no intention of ever using the sandbox.  If it's turned off then the implementation shouldn't have any extra synchronization cost over previous ONGNL, should it ?

There is also the issue of unexpected side-effects from the more-restrictive sandbox security manager running while any OGNL expression invocations are taking place ... which could break non-OGNL code running in the same JVM while it's in effect (in hard-to-track ways) ... unless I'm missing something.   :(

It seems like the _sandbox may be too far-reaching for its intended purpose_, but maybe I'm overlooking or misunderstanding.  :confused:

Maybe a sandboxed OGNL should be a parallel fork (so it can be swapped-in/swapped-out at the .jar level) ?  Or maybe a duplicate-but-separate `OgnlRuntimeSandboxed` that the library user decides at compile or runtime to use (instead of the standard `OgnlRuntime`) ?  :confused:

yasserzamani: 👍 Great catch! I'll move it outside synchronized block, thanks! - while it has a surrounding try-catch now, it's possible now to move it out :)

I think it doesn't break non-ognl code because of `isAccessDenied` check but you're right, it's currently JVM wide and currently I think it has to be because JDK's one is JVM wide - I understand your concern for example user has two web apps in same JVM and want sandbox for one and don't want for another. To address this I guess we should include `ProtectionDomain` concept in `isAccessDenied` method.

Yes I understand it's complexity and huge side-effects but I wasn't able a better complete robust solution for "expression injection". However, I feel we will handle all at end via improving `isAccessDenied` method logic - this PR gets better every day :) And **as far as I know OGNL, I generally see it rational for OGNL to disable sensitive actions at all because in case of an unknown "expression injection" flaw in user app, the attacker has a powerful language and compiler in hand to do anything! @ ognl_community  right?**

No I think all of these are finally handle-able via a good `isAccessDenied` method logic :)

yasserzamani:
Hi all,

I believe now it's **ready for squash and merge** - it's enough -- less restriction (e.g. allow by annotation on user method) is wrong and I could find mortal scenarios for each of my ideas.

P.S.
Now, still there are only 7 (of 1842) test fails in Struts but they are not sandbox issues but are awesome catches by sandbox! - Most of them are because of https://github.com/jkuhnert/ognl/blob/d95a4ef9dc8312e57eefd68c4ae2b782d61c3442/src/java/ognl/OgnlRuntime.java#L2213 The sandbox is right; on a successful "Expression Injection" attack, the attacker is able to use https://github.com/jkuhnert/ognl/blob/d95a4ef9dc8312e57eefd68c4ae2b782d61c3442/src/java/ognl/OgnlRuntime.java#L2183 which is mortal! (@harawata , it's OGNL issue itself)
However, [I think it's not a good idea to grant `accessDeclaredMembers` permission](https://docs.oracle.com/javase/7/docs/technotes/guides/security/permissions.html):

> Warning: Extreme caution should be taken before granting this permission to code, for it provides access to the declared members of a class.

What do you think? Does OGNL really need access to user non-public methods and fields?! I don't think, as far as I know OGNL. (FYI I found 15 occurrences of `Declared` word in OGNL codes)

harawata:
Hi @yasserzamani ,

> Does OGNL really need access to user non-public methods and fields?!

I think it is totally valid as an OGNL feature.
MyBatis uses it, at least. :)

I seem to be the only participant who is not from Struts team, so please allow me to remind you that the protection against expression injection attack is useful only for libraries/frameworks that pass unknown users' input to OGNL (which, IMHO, developers should avoid in general).

Like `Runtime.exec()`, OGNL is very powerful and that is not wrong nor vulnerable by itself.
It just requires developers to use it safely.
In that regard, I found @JCgH4164838Gh792C124B5 's [ideas](https://github.com/jkuhnert/ognl/pull/69#discussion_r279236265) quite interesting.

yasserzamani:
So I think following pseudo code could be more nice to honor user if a security manager is set:

```java
try {final Method[] methods = c.getDeclaredMethods(); }
catch (SecurityException se) {final Method[] methods = c.getMethods(); }
```

Anyway, **IMHO it's OK to merge this PR please** - It's disabled by default with somehow zero side effects. Users who wish, can try `-Dognl.security.manager` JVM option on their test environment and run their integration and UI tests. Then they will be able to fix any issue and improve their security and try it in their production environment. It's nice, isn't? :)

P.S. Struts complex showcase app now is able to operate correctly with this PR. Still just 7 (less than 1%) of core tests fail but are expected to fail.

JCgH4164838Gh792C124B5:
It appears the dynamic install/uninstall of a SecurityManager instance by **the sandbox has a Java-wide impact** (application / application server wide).  So ... enabling **the OGNL sandbox could cause application-server wide failures to occur** (even outside of the application using OGNL) whenever the OGNL sandbox is active (i.e. due to any OGNL method invocation at any given instant).

_Unfortunately_, I think the sandbox still represents too great a "_breakage risk_", since _the sandbox cannot be isolated to only the OGNL method invocation calls_ ...   :disappointed:

yasserzamani:
But my isAccessDenied method returns true only if OGNL_SANDBOX_CLASS_NAME equals c.getName() i.e. only when ognl.security.UserMethod is inside call stack so I think it is isolated to only the OGNL method invocation calls. Isn't it?

JCgH4164838Gh792C124B5:
OK sorry about that ... the `isAccessDenied()` method looks like it should only return true when an OGNL method call is on the stack ... but I had to read the code several times before seeing that.  _The implementation is sophisticated enough_ that I can easily miss or misunderstand things.  ;)

Putting that concern aside, after reading the most recent revision a few questions arose:
1) `OgnlSecurityManager` constructor doesn't call super() - was that intentional ?
2) When it is _active_ (_with a parent **SecurityManager**_), it doesn't doesn't define a _full set of methods to call the parent's implementation_.  Might that not cause different behaviour for non-OGNL calls during those time frames (_possibly breaking things elsewhere_ since the default behaviour might be different than the parent) ?
3) Does `isInsideClassLoader` have to be `Boolean`, or would a _primitive_ `boolean` suffice ?

Still _a little uneasy_ about two things:
1) Multiple install/uninstall of SecurityManager instances over time as it does not seem to be a typical usage pattern (it seems that usually a sm gets installed and stays installed until the JVM ends).  Beyond the synchronization overhead, numerous installs/uninstalls might lead to some unexpected behaviour in the JVM.
2) Having to call `System.getProperty("ognl.security.manager")` for every method invocation (to determine if sandbox enabled) every time may be overly expensive (_especially if another SecurityManager is installed but not using the OgnlSecurityManager_ each call involves a permission check).  One potential way to minimize the check cost would be to have local state to check (like a boolean or null check) instead ...

What to you think ?

lukaszlenart:
After trying to understand all those changes, wouldn't be better to provide a custom security policy file that can be used by the Struts users if they want to have a fully secure application?

Instead of forcing others (non-Struts users) to cope with Ognl Security Manager, we could easily show an alternative solution to the Struts users.

yasserzamani:
Hi,

> The implementation is sophisticated enough that I can easily miss or misunderstand things. ;)

Yes, reviewing other works it is, thanks for it! - you now know what I feel when I see yours ones :)

> 1. OgnlSecurityManager constructor doesn't call super() - was that intentional ?

I think if the super class constructor is public and has no arguments Java automatically calls it for you. If it has arguments you'll get an error.

> 2. When it is active (with a parent SecurityManager), it doesn't doesn't define a full set of methods to call the parent's implementation. Might that not cause different behaviour for non-OGNL calls during those time frames (possibly breaking things elsewhere since the default behaviour might be different than the parent) ?

As far as I could see, every checkXXX, calls and relies on checkPermission result. I found it enough to only override it and it does nothing for non-OGNL user methods. However, I don't think anyone use already a security manager with OGNL as it doesn't work with.

> 3. Does isInsideClassLoader have to be Boolean, or would a primitive boolean suffice ?

I use null for not still known value. When it gets false or true, isInsideClassLoader==null is false and execution doesn't enter inside if statement because when determined, it shouldn't be evaluated or changed again.

> Still a little uneasy about two things:
> 1. Multiple install/uninstall of SecurityManager instances over time as it does not seem to be a typical usage pattern (it seems that usually a sm gets installed and stays installed until the JVM ends). Beyond the synchronization overhead, numerous installs/uninstalls might lead to some unexpected behaviour in the JVM.

Me too. Unfortunately I probed a lot but couldn't find any doc about these and custom security manager and etc; Those what you mentioned are for performance. I think it's my best attempt. Unit tests work. I think it doesn't harm anyone to merge this option and see what happens and what is users opinions :)

Anyway I think it isn't good to stop adventure and say it might break something without any POC, then we cannot introduce anything - I have done my best attempt and always can fix it in future :)

> 2. Having to call System.getProperty("ognl.security.manager") for every method invocation (to determine if sandbox enabled) every time may be overly expensive (especially if another SecurityManager is installed but not using the OgnlSecurityManager each call involves a permission check). One potential way to minimize the check cost would be to have local state to check (like a boolean or null check) instead ...

It is to enable user to enable/disable it lazily runtime. I need it also for testing :( However, I don't think anyone use already a security manager with OGNL as it doesn't work with. But I would like to deliver it as is and always can fix it in future.

Regards.

yasserzamani:
Hi,

> After trying to understand all those changes, wouldn't be better to provide a custom security policy file that can be used by the Struts users if they want to have a fully secure application?
> Instead of forcing others (non-Struts users) to cope with Ognl Security Manager, we could easily show an alternative solution to the Struts users.

As far as I could see, a policy file is a list of ProtectionDomain. A ProtectionDomain is a (CodeSource, Permissions) tuple e.g. ('somepath/webapps/myWebApp/WEB-INF/lib/ognl.jar', NoPermission). i.e. you only can specify that a *whole* code source e.g. ognl.jar has which permissions. So each permission you have to grant to ognl.jar (e.g. classLoading, suppressAccessChecks and etc) to make it working, user methods also will have those sensitive permissions because all of them are in one ProtectionDomain. If you don't grant them then ognl.jar itself stops working. If you grant them then you have to also grant them to user methods.

Anyway, yes. It was better if OGNL was able to work with a default JDK security manager. If so, then I could drop OgnlSecurityManager class but rest code (codes about protection domains, permissions ad doPrivileged) was still needed.

Regards.

lukaszlenart:
Ok, the last call, if no objections from the others I will merge/squash this and release as a new 3.1.x version.

yasserzamani:
Hi @JCgH4164838Gh792C124B5 

This repository and project isn't inside ASF and from other old files, it seems this doesn't force such license header.

Once the PR is merged, I'm going to compile an article in Struts site as an experimental security advice that they can try this feature in their test environment and see if their UI, integration and unit tests pass then they can try in production environment. I'll request them report any issue or feedback please. Then we will see how much this feature is used and liked at all.

lukaszlenart:
This project uses Apache License 2.0, a dedicated header is not required.

lukaszlenart:
🎉

lukaszlenart:
I will release a new 3.1.x version right now

yasserzamani:
Thanks for confidence in my work :tada: Now it's time to adapt Struts side and document this feature there :)

PR #70, opened 2019-03-31
Proposed minor cleanup change to 3.1.x branch

JCgH4164838Gh792C124B5:
Proposed minor cleanup change to 3.1.x branch:
- `OgnlRuntime.invokeMethod()`:
  - Eliminate a couple of unused assignments, reduce number of `get()` calls to `ConcurrentHashMap` instances by 2.
- `OgnlRuntime._getGetMethod()`: Removed one whitespace line.
- `OgnlRuntimeTest`: Added some execution time tracking and output to runTest().
- `OgnlRuntimeTest`: Added two additional tests involving a larger set of test classes (10) and calling two methods per class.

### DIFF
--- a/src/java/ognl/OgnlRuntime.java
+++ b/src/java/ognl/OgnlRuntime.java
@@ -819,49 +819,55 @@ public class OgnlRuntime {
     public static Object invokeMethod(Object target, Method method, Object[] argsArray)
             throws InvocationTargetException, IllegalAccessException
     {
-        boolean syncInvoke = false;
-        boolean checkPermission = false;
+        boolean syncInvoke;
+        boolean checkPermission;
+        Boolean methodAccessCacheValue;
+        Boolean methodPermCacheValue;
 
         // only synchronize method invocation if it actually requires it
 
         synchronized(method) {
-            if (_methodAccessCache.get(method) == null) {
+            methodAccessCacheValue = _methodAccessCache.get(method);
+            if (methodAccessCacheValue == null) {
                 if (!Modifier.isPublic(method.getModifiers()) || !Modifier.isPublic(method.getDeclaringClass().getModifiers()))
                 {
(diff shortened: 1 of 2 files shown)


JCgH4164838Gh792C124B5:
Hello OGNL Team.

The proposed changes came to mind after reviewing the runtime module's code again.  Reducing the number of calls to the synchronized hashmaps seemed like a good idea, and appears possible.

The two new unit tests take a while to run, but overall still completes the build under 2 minutes.  The execution timeout sysouts in the tests just provide some historical information about execution time that should make detecting wild swings in performance easier to see in the build results.

After running many runs with the proposed code and old code I'm pretty confident its execution time is more consistent (if not significantly faster in tests, should perform better than the old code when the number of `ConcurrentHashMap` entries grows larger).

What do you think ?

_Note_:  During one test/build run with the _original `OgnlRuntime` code_ I experienced an `AssertionFailedError` in `runtest()` (during the Assert.assertArrayEquals()) call).  However I couldn't reproduce it even after raising the number of iterations from 10 000 000 to 100 000 000 and running the suite multiple times.  Since I couldn't reproduce it I'm not sure where it might have come from ...

lukaszlenart:
Looks good to me, @harawata @kazuki43zoo @danielfernandez do you have any objections? If not I will merge this and push a new release.

@JCgH4164838Gh792C124B5 thanks a lot!

harawata:
I don't see any harm. 👍

lukaszlenart:
LGTM 👍

lukaszlenart:
OGNL 3.1.23 is under way to the Central

JCgH4164838Gh792C124B5:
Hello @lukaszlenart 

Do you want an equivalent PR for 3.2.x (or do you prefer cherry-picking) ?

lukaszlenart:
Cherry-picks are fine, thank you :)

PR #71, opened 2019-03-31
Proposed enhancement for the 3.1.x branch (method deny list, related)

JCgH4164838Gh792C124B5:
Proposed enhancement for the 3.1.x branch:  
[Updated] - Introduced convention for specialized processing (method deny list is an implementation)
- Inspired by Y. Zamani's proposed sandboxing idea in PR#69.
- Provide an opt-in method deny list for OGNL.
- By default the method deny list is empty (no impact to existing usage).
- A method to add an entry (one-at-a-time) is provided (for custom configuration).
- Two standard deny list preparation methods are also provided.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -10,7 +10,7 @@
     <groupId>ognl</groupId>
     <artifactId>ognl</artifactId>
     <packaging>jar</packaging>
-    <version>3.1.23-SNAPSHOT</version>
+    <version>3.1.24-SNAPSHOT</version>
     <name>OGNL - Object Graph Navigation Library</name>
     <description>OGNL - Object Graph Navigation Library</description>
 
--- a/src/java/ognl/ListPropertyAccessor.java
+++ b/src/java/ognl/ListPropertyAccessor.java
@@ -244,7 +244,7 @@ public class ListPropertyAccessor extends ObjectPropertyAccessor implements Prop
                 
                 if (m != null || !context.getCurrentType().isPrimitive())
                 {
-                    System.out.println("super source setter returned: " + super.getSourceSetter(context, target, index));
+                    //System.out.println("super source setter returned: " + super.getSourceSetter(context, target, index));
                     return super.getSourceSetter(context, target, index);
                 }
 
--- a/src/java/ognl/OgnlMethodBlocker.java
+++ b/src/java/ognl/OgnlMethodBlocker.java
@@ -28,26 +28,21 @@
 // THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
 // DAMAGE.
 // --------------------------------------------------------------------------
-package ognl.delegate;
+package ognl;
 
-import ognl.OgnlRuntime;
 import ognl.enhance.OgnlExpressionCompiler;
 
 import java.io.File;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.io.PrintStream;
-import java.lang.reflect.AccessibleObject;
 
-import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
-import java.lang.reflect.Modifier;
 
 import java.util.Map;
 import java.util.Properties;
 import java.util.concurrent.ConcurrentHashMap;
-import ognl.Node;
-import ognl.OgnlContext;
+
 
 /**
  * Utility class used by internal OGNL API to do delegate (alternate) OgnlRuntime processing.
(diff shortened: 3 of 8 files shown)


JCgH4164838Gh792C124B5:
Hello OGNL Team.

On the surface the changes in this PR appear to work as intended (the previous unit tests and new unit test all pass).  A build for Struts 2.5.21-SNAPSHOT completed with the PR changes, and the showcase application appears to function normally with it as well.

It is nowhere near as sophisticated as PR#69, and the same considerations for performance will need to be considered for it as well.

JCgH4164838Gh792C124B5:
Hello again.

After further reading and consideration, this PR has been updated to use a more modern and clearer naming for the method deny list/feature.

JCgH4164838Gh792C124B5:
This PR's implementation currently provides no mechanism to clear/empty entries in the deny list - _in theory_ that should make circumventing it more difficult.

The implementation could be leveraged by anything that uses OGNL (including Struts itself or an application using Struts).  There could also be synergy with the sandboxing idea @yasserzamani raised in PR#69 (whether at the OGNL level or at the Struts level).

yasserzamani:
Thank you but I think it's not needed. Please refer to [1] for the
reasoning.

[1] https://github.com/jkuhnert/ognl/pull/69#issuecomment-479468500

JCgH4164838Gh792C124B5:
Hello @yasserzamani.

Thanks for the comment and explanation in PR#69.  I think I understand the reasoning better because of that.  :)

One clarification - the idea for this PR is that instead of trying to limit "particular expressions", the mechanism provides a way to block invocation at the `Method` level (whether there's a security manager installed or not).  It does not provide the same level of guard/coverage of PR#69's idea, but since many "bad expressions" involve a subset of Java "system methods" (whether directly or through some call-chain in another library), it could provide a _reasonable level of protection at a reasonable cost_ (when used, _overhead should be similar to the existing permission/access cache mechanisms_).

Either way it was interesting to think about - and those thoughts arose in response to your PR.  :) :)

yasserzamani:
> It does not provide the same level of guard/coverage
> of PR#69's idea, but since many "bad expressions" involve a subset of
> Java "system methods" (whether directly or through some call-chain in
> another library)


No it doesn't work "through some call-chain in another library". We
already have same one in Struts. The problem is, say Struts has an
unknown "expression injection" flaw and some body will be able to inject
e.g. "new
some.third.party.package.SomeThirdPartyClass().someMethod('cat',
'/etc/pass')" where someMethod can execute commands. The above idea
doesn't block this because someMethod isn't subset of Java "system
methods" and OGNL cannot be aware about it's body (it only calls it via
reflection). It's impossible I think to discover and maintain all attack
vectors available in class path over time and adding them to deny list.
But JDK security model already provides it and in #69 blocks above
example when someMethod tries to execute a command. It blocks all
sensitive actions even when we have missed them (whether not discovered
at all or newly added over time).

JCgH4164838Gh792C124B5:
Thanks for the feedback :smile: (after seeing it and re-reading the statement you pointed out "through some call-chain in another library" ... I realized it wasn't as clear as I had intended ... so I attempted to improve it in an after-the-fact edit of the original comment).

The earlier comment has been edited (for clarification) in response to Y. Zamani's comment above.
As he clarified this PR's mechanism _cannot_ block any "currently unknown" or "future unknown" risk introduced in an OGNL expression.  

It can only deny a list of `Method` instances (which requires that you know about them ahead of time ... or at least be able to dynamically determine something at runtime and then add them).  However there may be value to block _direct_ invocation off some known risky `Method`(s) (or other specific known `Method`(s) for whatever reason, in a low-overhead fashion ... another "tool in the toolbox").

JCgH4164838Gh792C124B5:
Hello OGNL Team.

This PR has been refactored yet again to return to a single package (and removes the idea of "customizable delegation" due to fears it might be too easy to abuse).  

The latest revision adds support of an _optional_ limited security manager implementation that attempts to limit its impact to a "per-thread" level.  It can be enabled via JVM option (environment) or by explicit call.  It can be used together or independently of the method blocking mechanism as well.

Still not sure this PR provides a complete solution, but it might have some merit in terms of balanced options ...

lukaszlenart:
@JCgH4164838Gh792C124B5 sorry using this channel but could you send me an email or contact me on private in other way?

JCgH4164838Gh792C124B5:
Hello OGNL Team.

The current implementation has limitations on reads (current directory and below) and writes (any), which _will cause security exception failures for any log or system.out writes called during a method invocation sequence_ ... this _could be_ too great a limitation ...  :disappointed:

lukaszlenart:
I would stop here, this is a great work but I blocking writes is a huge blocker.

JCgH4164838Gh792C124B5:
Based on the commentary provided for this PR and the incorporation of the sandbox logic into 3.1.24, this PR does not appear to be needed at this time.
Closing this PR for now.

PR #72, opened 2019-05-19
Minor fix: Comment out missed debugging system.out.println call

JCgH4164838Gh792C124B5:
Minor fix:
- Comment out a system.out.println() call that appears to be
  debugging-related.
- All other system.out.println calls in OGNL are already commented out.

### DIFF
--- a/src/java/ognl/ListPropertyAccessor.java
+++ b/src/java/ognl/ListPropertyAccessor.java
@@ -244,7 +244,7 @@ public class ListPropertyAccessor extends ObjectPropertyAccessor implements Prop
                 
                 if (m != null || !context.getCurrentType().isPrimitive())
                 {
-                    System.out.println("super source setter returned: " + super.getSourceSetter(context, target, index));
+                    // System.out.println("super source setter returned: " + super.getSourceSetter(context, target, index));
                     return super.getSourceSetter(context, target, index);
                 }
 


JCgH4164838Gh792C124B5:
Hello OGNL Team.
Unless I'm mistaken, this PR should correct an unwanted system.out that appears to be intended for debugging.

src/java/ognl/ListPropertyAccessor.java
```diff
@@ -244,7 +244,7 @@ public String getSourceSetter(OgnlContext context, Object target, Object index)
                 
                 if (m != null || !context.getCurrentType().isPrimitive())
                 {
-                    System.out.println("super source setter returned: " + super.getSourceSetter(context, target, index));
+                    // System.out.println("super source setter returned: " + super.getSourceSetter(context, target, index));
```
lukaszlenart: Maybe throw it away?

JCgH4164838Gh792C124B5: I'm not sure if it should be thrown away, since there are a bunch of similar commented-out println() calls throughout OGNL ?

I think the statement (and similar ones elsewhere) are intended to be "uncommented locally" when debugging the "OGNL internals".  It's quicker to uncomment-and-recomment than have to re-create them ... but I could be mistaken.  :smile:

lukaszlenart: 👍

lukaszlenart:
LGTM 👍

lukaszlenart:
Would you cherry-pick this change to the master branch? Thanks in advance :)

PR #74, opened 2019-06-01
Proposed OGNL enhancement (alt. reflective access for JDK9+, enh. invocation control):

JCgH4164838Gh792C124B5:
Proposed OGNL enhancement:
- Implement mechanism to avoid "Illegal Reflective Access" warnings in JDK9+, with fallback to normal reflection.  Users can force usage of normal reflection in JDK9+ using a Java option (environment) flag if needed.
- Add interface to decouple AccessibleObject processing (easier support for future Java reflection changes).
- Add mechanism to check for JDK version (allows decision on access mechanism to use).
- Implement a stricter invocation mechanism by default to limit certain invocations.  Users can disable the stricter invocation using a Java option (environment) flag if needed.
- Adds additional unit tests for the changes (insofar as they can be tested).

### DIFF
--- a/src/java/ognl/AccessibleObjectHandler.java
+++ b/src/java/ognl/AccessibleObjectHandler.java
@@ -1,0 +1,51 @@
+//--------------------------------------------------------------------------
+//	Copyright (c) 1998-2004, Drew Davidson and Luke Blanshard
+//  All rights reserved.
+//
+//	Redistribution and use in source and binary forms, with or without
+//  modification, are permitted provided that the following conditions are
+//  met:
+//
+//	Redistributions of source code must retain the above copyright notice,
+//  this list of conditions and the following disclaimer.
+//	Redistributions in binary form must reproduce the above copyright
+//  notice, this list of conditions and the following disclaimer in the
+//  documentation and/or other materials provided with the distribution.
+//	Neither the name of the Drew Davidson nor the names of its contributors
+//  may be used to endorse or promote products derived from this software
+//  without specific prior written permission.
+//
+//	THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+//  "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+//  LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+//  FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+//  COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+//  INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+//  BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS
+//  OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED
+//  AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+//  OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
+//  THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
+//  DAMAGE.
+//--------------------------------------------------------------------------
+package ognl;
+
+import java.lang.reflect.AccessibleObject;
+
+/**
+ * This interface provides a mechanism for indirect reflection access processing
+ *   of AccessibleObject instances by OGNL.  It can be used to provide different
+ *   behaviour as JDK reflection mechanisms evolve.
+ *
+ * @since 3.1.24
+ */
+public abstract interface AccessibleObjectHandler
+{
+    /**
+     * Provides an appropriate implementation to change the accessibility of accessibleObject.
+     *
+     * @param accessibleObject
+     * @param flag
+     */
+    void setAccessible(AccessibleObject accessibleObject, boolean flag);
+}
--- a/src/java/ognl/AccessibleObjectHandler.java
+++ b/src/java/ognl/AccessibleObjectHandler.java
@@ -1,33 +1,21 @@
-//--------------------------------------------------------------------------
-//	Copyright (c) 1998-2004, Drew Davidson and Luke Blanshard
-//  All rights reserved.
-//
-//	Redistribution and use in source and binary forms, with or without
-//  modification, are permitted provided that the following conditions are
-//  met:
-//
-//	Redistributions of source code must retain the above copyright notice,
-//  this list of conditions and the following disclaimer.
-//	Redistributions in binary form must reproduce the above copyright
-//  notice, this list of conditions and the following disclaimer in the
-//  documentation and/or other materials provided with the distribution.
-//	Neither the name of the Drew Davidson nor the names of its contributors
-//  may be used to endorse or promote products derived from this software
-//  without specific prior written permission.
-//
-//	THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
-//  "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
-//  LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
-//  FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
-//  COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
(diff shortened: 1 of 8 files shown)


JCgH4164838Gh792C124B5:
Hello @lukaszlenart .

This PR came about following the Jira report of ""Illegal Reflective Access" warnings in JDK 11.  After some research (as noted in the code comments) it turns out there is a strategy to avoid those warnings - which this PR attempts to achieve.

With the above change it seemed prudent to introduces a basic "stricter" invocation mode alongside it.  No micro-benchmarking was performed, but according to some nanotime sysouts during development, the check takes on average 0.01 milliseconds (and only 2-3 out of 1500+ calls in the regression run reporting as high as 0.49 milliseconds using an i5-5200u).

All of the new behaviour can be disabled (independently) via a separate Java option flag for users/application that need or desire to do so.

Using the snapshot jar for this PR, the `Struts 2.5.21-SNAPSHOT` build completes with `100% tests passing` as well.  :smile:

Please let me know what you think.

src/java/ognl/AccessibleObjectHandler.java
```diff
@@ -0,0 +1,51 @@
+//--------------------------------------------------------------------------
+//	Copyright (c) 1998-2004, Drew Davidson and Luke Blanshard
```
lukaszlenart: I would either put your credits or use a standard ASL 2.0 header instead - this code cannot be credited by the above authors :)

src/java/ognl/AccessibleObjectHandlerJDK9Plus.java
```diff
@@ -0,0 +1,236 @@
+//--------------------------------------------------------------------------
+//	Copyright (c) 1998-2004, Drew Davidson and Luke Blanshard
```
lukaszlenart: Same here

src/java/ognl/AccessibleObjectHandlerJDK9Plus.java
```diff
@@ -0,0 +1,236 @@
...
+    private static final Object _unsafeInstance = instantiateUnsafeInstance(_clazzUnsafe);
+    private static final Method _unsafeObjectFieldOffsetMethod = instantiateUnsafeObjectFieldOffsetMethod(_clazzUnsafe);
+    private static final Method _unsafePutBooleanMethod = instantiateUnsafePutBooleanMethod(_clazzUnsafe);
+    private static final Field _accessibleObjectOverrideField = instantiateAccessibleObjectOverrideField();
+    private static final long _accessibleObjectOverrideFieldOffset = determineAccessibleObjectOverrideFieldOffset();
+
+    /**
+     * Package-accessible constructor
+     */
+    AccessibleObjectHandlerJDK9Plus() {}
```
lukaszlenart: Why not private?

src/java/ognl/AccessibleObjectHandlerPreJDK9.java
```diff
@@ -0,0 +1,59 @@
+//--------------------------------------------------------------------------
+//	Copyright (c) 1998-2004, Drew Davidson and Luke Blanshard
```
lukaszlenart: Same here

src/java/ognl/DefaultMemberAccess.java
```diff
@@ -47,6 +47,14 @@
  */
 public class DefaultMemberAccess implements MemberAccess
 {
+    /*
+     * Assign an accessibility modification mechanism, based on Major Java Version.
+     *   Note: Can be override using a Java option flag {@link OgnlRuntime#USE_PREJDK9_ACESS_HANDLER}.
+     */
+    private static final AccessibleObjectHandler _accessibleObjectHandler =
+            (OgnlRuntime.isJdk9Plus() && !OgnlRuntime.getUsePreJDK9AccessHandlerValue()) ? new AccessibleObjectHandlerJDK9Plus() :
+            new AccessibleObjectHandlerPreJDK9();
```
lukaszlenart: Would you mind wrapping this whole logic into a method?

src/java/ognl/AccessibleObjectHandler.java
JCgH4164838Gh792C124B5: I wasn't sure of the best header to use (given the differences in the source and license.txt file),  
I've attempted to use a slight modification of the ASF header from some Struts files that seemed to fit the circumstances reasonably well.  
All three files have been updated with that header - let me know if it looks OK now,

src/java/ognl/AccessibleObjectHandlerJDK9Plus.java
JCgH4164838Gh792C124B5: That was intended to restrict creation of instances to the package, so that only classes within the main package could construct an instance.
With the latest change the constructors are now private and instead a package-level generator method is called to produce an instance (which serves the same purpose, but is a bit easier to read ... plus they just return the generic type reference).

src/java/ognl/AccessibleObjectHandlerPreJDK9.java
JCgH4164838Gh792C124B5: Please see above response.

src/java/ognl/DefaultMemberAccess.java
JCgH4164838Gh792C124B5: I tried a compromise in the latest push.  There is now a single method to indicate use of the JDK9Plus handler, combined with the two generator methods.
  
Using that combination in both `OgnlRuntime` and `DefaultMemberAccess`, along with static initializer blocks (which is more in line with other initialization structures in the package), looks a bit cleaner ... I hope. :smile:

yasserzamani:
Excuse me, there are clever codes, but basically doesn't look good to me. Java 9 tries to learn us to deprecate and then fix our not secure codes, but this PR makes things worse :) - those are temporal-working not maintainable and readable complex hacks via `Unsafe`. Java 9 doesn't mean this by that warning, I think :). Additionally, I guess Java 9 may already aware on such hacks and also will produce a warning or an exception on such behaviors similarly.

Instead, I think we should deprecate such things and fix root causes and issues in next versions. For example, when I experienced Struts tests with `--illegal-access=debug`, I realized almost all of them are because `java.util.HashMap$Node.(getKey/getValue)` where method is public but class is not! I realized that the root cause/issue is: **OGNL isn't able to find appropriate/best method correctly** - in this case, when user e.g. request "key" on a HashMap entry, OGNL must call it's interface method, `Map.Entry.(getKey/getValue)` which is the most best and accessible (doesn't need setAccessible) method. But OGNL incorrectly prefers `java.util.HashMap$Node.(getKey/getValue)` which needs setAccessible :(

JCgH4164838Gh792C124B5:
Hello.
 I _**agree** that avoiding the need to change accessibility is a good idea when it is possible_.  That being said, there will _probably_ still remain use cases where OGNL may need to change accessibility in order to provide the same-level-of-service it can today (for existing applications using the library).  

Until the future of post-JDK9 reflection use in later JDKs is clear, it seems prudent to "keep options open" for now ... and adapt as newer JDK releases arrive.  It may even be a question as to whether OGNL should be able to change accessibility by default ... which could also be controlled by an environment flag (it seems that I like control flags :wink:).

If the _**preference** is to **not** "mask" the reflection warnings by default_ in JDK9+, this PR can be modified so that it just continues to use normal reflection as a default.  Instead, _the environment flag could be used to enable the alternate mechanism for JDK9+ only_ (so a user is making a conscious decision to enable).  That seems to be a _balanced_ approach, so I'll update the PR to do that instead.

yasserzamani:
Could you please run Struts tests with this PR, java 9 and 11, and *--illegal-access=deny* JVM option? (Please note that this jvm option didn't work for me in Maven options, you should add this to jvm options at first place before jvm start)

JCgH4164838Gh792C124B5:
I don't currently have a JDK9 installed, so I only tested with JDK 11.  Unfortunately it does not appear to be possible to run the Struts build (regression) with "illegal-access=deny" as other test libraries need that level of reflection and fail instead (e.g. Spring proxy enhancer).

I was able to _manually_ test the Struts Showcase application under JDK11 with "--illegal-access=deny" enabled and this PR's version of OGNL.  Combined with "-Dognl.UseJDK9PlusAccessHandler=false" (or omitted), the "Option Transfer Select UI Example" fails.  Combined with "-Dognl.UseJDK9PlusAccessHandler=true" that same UI Tag example succeeds.

The manual test was with the latest revision of the PR (where the alternate access mechanism is opt-in, and only for JDK9+).

yasserzamani:
You're right. I tested also. This was not expected for me that Java allows `theUnsafe.setAccessible(true)` even with `--illegal-access=deny` but anyway your work works :) 👍

JCgH4164838Gh792C124B5:
This PR will be updated (hopefully soon) after merging with 3.1.24, making any necessary changes and confirming functionality.

lukaszlenart:
Should I merge this or #75 takes precedence over it?

JCgH4164838Gh792C124B5:
Hello @lukaszlenart .

I consider this PR is independent as it addresses different issues.  Beyond the JDK9+ alternate accessibility mechanism, this PR includes the "stricter" invocation control idea plus the new "forceDisableOnInit" flag for the optional 3.1.24 security manager.

The PR #75 aims to avoid some of the "illegal reflective access" scenarios for getters/setters, but doesn't address the full range of concerns this PR does.

I think this PR could be merged first, with #75 considered next,

lukaszlenart:
LGTM 👍

PR #75, opened 2019-06-10
OGNL improve getter/setter detection algorithm:

JCgH4164838Gh792C124B5:
OGNL improve getter/setter detection algorithm:
- Add detail to isDefaultMethod() method comment.
- Add a new isNonDefaultPublicInterfaceMethod() method.
- Modify collectAccessors() method to consider public interface methods in addition to default methods.
- Modify _getGetMethod() to prefer the first valid public getter over other getters, then the first valid public non-Default interface getter and lastly the first getter of any kind.
- Modify _getSetMethod() to prefer the first valid public setter over other setters, then the first valid public non-Default interface setter and lastly the first setter of any kind.

### DIFF
--- a/src/test/java/ognl/OgnlRuntimeTest.java
+++ b/src/test/java/ognl/OgnlRuntimeTest.java
@@ -316,4 +316,34 @@ public class OgnlRuntimeTest {
                     (cumulativelRunTestElapsedNanoTime / (1000 * 1000 * 1000)) / totalNumberOfRunTestRuns + " s)");
         }
     }
+
+    /**
+     * Test OgnlRuntime value for _useFirstMatchGetSetLookup based on the System property
+     *   represented by {@link OgnlRuntime#USE_FIRSTMATCH_GETSET_LOOKUP}.
+     */
+    @Test
+    public void testUseFirstMatchGetSetStateFlag() {
+        // Ensure no exceptions, basic ouput for test report and sanity check on flag state.
+        final boolean defaultValue = false;          // Expected non-configured default
+        boolean optionDefinedInEnvironment = false;  // Track if option defined in environment
+        boolean flagValueFromEnvironment = false;    // Value result from environment retrieval
+        try {
+            final String propertyString = System.getProperty(OgnlRuntime.USE_FIRSTMATCH_GETSET_LOOKUP);
+            if (propertyString != null && propertyString.length() > 0) {
+                optionDefinedInEnvironment = true;
+                flagValueFromEnvironment = Boolean.parseBoolean(propertyString);
+            }
+        } catch (Exception ex) {
+            // Unavailable (SecurityException, etc.)
+        }
+        if (optionDefinedInEnvironment) {
+            System.out.println("System property " + OgnlRuntime.USE_FIRSTMATCH_GETSET_LOOKUP + " value: " + flagValueFromEnvironment);
+        } else {
+            System.out.println("System property " + OgnlRuntime.USE_FIRSTMATCH_GETSET_LOOKUP + " not present.  Default value should be: " + defaultValue);
+        }
+        System.out.println("Current OGNL value for Use First Match Get/Set State Flag: " + OgnlRuntime.getUseFirstMatchGetSetLookupValue());
+        Assert.assertEquals("Mismatch between system property (or default) and OgnlRuntime _useFirstMatchGetSetLookup flag state ?",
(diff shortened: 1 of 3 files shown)


JCgH4164838Gh792C124B5:
Hello OGNL Team.

This PR attempts to address a behaviour that @yasserzamani identified when reviewing PR#74 and illegal reflective access warnings running Struts 2.5.x under JDK 11.

Currently OGNL 3.1.x picks the _first_ getter/setter that it can find, even if the method is non-public or the declaring class is non-public.  The updated logic in this PR attempts to find the _first publicly accessible_ getter/setter method (both method and declaring class public) if possible, considering methods from the class and interfaces implemented by the class.

After a few failed attempts, this implementation _seems_ to work.  It passes the regression tests for OGNL as well as the regression build for Struts 2.5.21-SNAPSHOT when used as part of the build. 

Running the 2.5.21. Showcase application under JDK 11 with this PR's implementation also seems to avoid the illegal reflective access warnings on the console too.  :smile:

Since the logic for choosing the getter/setter has changed, _this change presents some risk_ - since it could choose differently than previous OGNL 3.1.x versions.  However, choosing a publicly accessible getter/setter seems like better practice if possible to do so.

Please advise what you think about the proposed change, and whether the logic seems sound or not.

yasserzamani:
osm 😮  LGTM 👍

yasserzamani:
And consequently it seems #74 isn't essential anymore (?)

JCgH4164838Gh792C124B5:
Thanks for taking a look.  :smile:
I think the changes for #74 are still desirable for OGNL 3.1.x in general (to avoid breaking existing application usage and keeping options open for JDK11). 
At least this PR should help "head in the direction" of minimizing the need for the accessibility modification feature, at least for getter/setter methods.

JCgH4164838Gh792C124B5:
This PR will be updated (hopefully soon) after merging with 3.1.24, making any necessary changes and confirming functionality.

JCgH4164838Gh792C124B5:
Hello.

I'm wondering if this PR _also_ needs a **control flag** to decide whether to use the old or new lookup logic..  Although the new logic _should choose a "better" match_ it's possible that existing applications could encounter unexpected behaviour as a result (due to a different getter or setter chosen with new vs. old logic).

An environment flag (or static library method) allowing users to control whether to use the new or old lookup logic would limit that particular risk.

The PR can be adjusted to add such a control mechanism (or any other changes that people think should be made).

src/test/java/ognl/OgnlRuntimeTest.java
```diff
@@ -316,4 +316,34 @@ public void testPerformanceMultipleClassesMultipleMethodsMultipleThreads() throw
...
+            final String propertyString = System.getProperty(OgnlRuntime.USE_FIRSTMATCH_GETSET_LOOKUP);
+            if (propertyString != null && propertyString.length() > 0) {
+                optionDefinedInEnvironment = true;
+                flagValueFromEnvironment = Boolean.parseBoolean(propertyString);
+            }
+        } catch (Exception ex) {
+            // Unavailable (SecurityException, etc.)
+        }
+        if (optionDefinedInEnvironment) {
+            System.out.println("System property " + OgnlRuntime.USE_FIRSTMATCH_GETSET_LOOKUP + " value: " + flagValueFromEnvironment);
```
sepe81: IMO it would be better to use logging instead of sysout even for testing.

JCgH4164838Gh792C124B5: Hi. 
Given the nature of this library, having tests output to sysout seems sufficient for now (AFAIK it's the only mechanism that has been available up until now, and the results are clearly visible in the Maven test run output).
Maybe someone will introduce a testing-only logging replacement for the sysouts in the tests in a future PR.  :smile:

lukaszlenart:
@JCgH4164838Gh792C124B5 conflicts :(

JCgH4164838Gh792C124B5:
Fixed the merge conflicts but now Travis CI complains about usage of source level 1.5. 😕 
Will change source/target to 1.6 (same as Yasser did for #79) and see if that makes Travis CI happy.

JCgH4164838Gh792C124B5:
Hello @lukaszlenart .

Manual resolution of merge conflicts appears to have worked. :smile:
A follow-up commit to use source/target 1.6 was required to get Travis CI checks to complete (could be set back to 1.5 if needed, though - compiles fine locally with JDK 7).

lukaszlenart:
LGTM 👍

Issue #34, opened 2017-03-02
OgnlRuntime.clearCache() Doesn't Clear All Caches


chrisohiou57:
The OgnlRuntime.clearCache() method does not empty the cacheSetMethod and cacheGetMethod caches. The inability to clear this cache is leading to the following error:

java.lang.IllegalArgumentException: object is not an instance of declaring class

It appears that the issue occurs when classes are reloaded by a different classloader. If we had the ability to clear the caches mentioned above we believe it would fix our issue.

In our case we are using Spring Boot Dev Tools. When the application restarts and classes are reloaded we have no way to clear the caches mentioned above.

Relates to [OGNL-257](https://issues.apache.org/jira/browse/OGNL-257)

Issue #42, opened 2017-12-12
Log flood with "Two methods with same score(0)"


rsandoz:
Our logs are being flooded with the message:
Two methods with same score(0): "public java.lang.String myPackage.MyClass.myMethod(myPackage.MyModelClass,int)" and "public abstract java.lang.String myPackage.MyClassAbstract.myMethod(java.lang.Object,int)" please report!

We are on the latest version of struts (struts2-core-2.5.14.jar) which uses ognl-3.1.15.

This issue seems to come from OgnlRuntime.java at line 1430.

I grepped on future versions of OgnlRuntime.java from your github for 'abstract' ignoring case and found code in 3.1.16 and 3.2.3, and 3.2.4.   I tried applying the logic from these code fixes in a locally compiled OgnlRuntime, but the problem did not go away.  I don't completely understand your score calculation, but some of the other flags (isInterface, etc) seem to preclude our implementation.

The only thing I could do to workaround this was to put above this sysout line:
if (!Modifier.isAbstract(m.getModifiers()))

Our implementation is:
<<MyClassAbstract.java>>
package myPackage;
public abstract class MyClassAbstract <E> {
    public abstract String myMethod (E element, int i);

<<MyClass.java>>
package myPackage;
public class MyClass extends MyClassAbstract < MyModelClass > {
    public String myMethod (MyModelClass myModel, int i) {

Please advise.

lukaszlenart:
You can try to use [OGNL 3.1.16](https://github.com/jkuhnert/ognl/blob/master/docs/VersionNotes.md#release-notes---version-3116-324)

rsandoz:
That was one of the first things we tried.  It didn't work.  Still had the log flood with said message.  

I actually compiled OgnlRuntime from 3.1.15 and put the class in my classpath ahead of the one baked into the jar.  In doing so, I could comment out the message, which I prefer not to do.

I tried incorporating just the abstract logic from future revisions in my custom copy of 3.1.15 OgnlRuntime.java and could not get it.

[...]

PR #77, opened 2019-06-29
Cache clearing enhancement (clear additional cache, synchronization fix):

JCgH4164838Gh792C124B5:
Attempt to resolve issue #34:
- Add a clearAdditionalCache() method to clear some additional cache elements.
Using a separate method as changing the behaviour of clearCache() could negatively impact existing usage.
- Added missing synchronization within clearCache() for many elements that should have synchronization for modifications.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -121,6 +121,9 @@
                     <archive>
                         <compress>true</compress>
                         <index>true</index>
+                        <manifestEntries>
+                            <Automatic-Module-Name>ognl</Automatic-Module-Name>
+                        </manifestEntries>
                     </archive>
                 </configuration>
             </plugin>
--- a/pom.xml
+++ b/pom.xml
@@ -78,11 +78,12 @@
 
         <plugins>
             <plugin>
+                <!-- Raised source/target levels to 1.6 as Travis CI now rejects 1.5 -->
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-compiler-plugin</artifactId>
                 <configuration>
-                    <source>1.5</source>
-                    <target>1.5</target>
+                    <source>1.6</source>
+                    <target>1.6</target>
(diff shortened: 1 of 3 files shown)


JCgH4164838Gh792C124B5:
Hello OGNL Team.

After reviewing open issues, decided to try to replicate issue #42 and was able to confirm that it still exists.  A new test confirmed the syserr output was occurring for the reported scenario but the actual call by OGNL worked just fine.

Looking at the code and "forcing both paths" (abstract and concrete) the proper result was returned in both cases.  So the `OgnlRuntime.findBestMethod()` in this PR has now been updated with checks so that it will not output the error anymore unless _both_ methods being compared have the _same "concreteness" or "abstractness"_ - since that would definitely be a weird state that should be reported.

JCgH4164838Gh792C124B5:
I also encountered some weird behaviour for some of the Security Sandbox tests when performing interactive debugging for the last update to this PR. 

Occasionally one of the sandbox tests `TestOgnlRuntime.test_Call_Method_In_JDK_Sandbox_Privileged()`)
failed due to the cause being a `NullPointerException` instead of the expected type.

Running the test module outside of the debugger never produced failures, so it's unclear if it was just interference by the IDE debugger or some underlying condition in the logic.  Just wanted to leave a note for future reference in case an issue pops up.

lukaszlenart:
@JCgH4164838Gh792C124B5 conflict :(

JCgH4164838Gh792C124B5:
Hello @lukaszlenart .

Manual resolution of merge conflicts completed (but might conflict if #75 is accepted/merged).
Additional commit to use source/target 1.6 was required to get Travis CI checks to complete (just like in #75).

It should be easier for others to review now, at least. :smile:

lukaszlenart:
LGTM 👍

PR #84, opened 2019-09-29
Backport PR#78 from 3.2.x to 3.1.x (cherry-pick)

JCgH4164838Gh792C124B5:
Backport PR#78 from 3.2.x to 3.1.x (cherry-pick) - Merge pull request #78 from peteruhnak/master.

Compare non-comparable objects by equals only

(cherry picked from commit 937fd071b99a7da1230a64d724aa72b96e1fff68)

### DIFF
--- a/src/java/ognl/OgnlOps.java
+++ b/src/java/ognl/OgnlOps.java
@@ -128,9 +128,9 @@ public abstract class OgnlOps implements NumericTypes
 
         if (object1 == object2) {
             result = true;
-        } else {
-            if ((object1 != null) && object1.getClass().isArray()) {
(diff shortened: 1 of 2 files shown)


JCgH4164838Gh792C124B5:
**Note**:  Projects using OGNL 3.1.x may be impacted due to no longer receiving IllegalArgumentException for certain expressions.  Some expressions that used to fail evaluation may now succeed (with false return) or return a differeent result with the new logic.

_It may be unlikely to impact users_ but a similar comment exists in the original PR for 3.2.x, reiterated here for future reference (in case someone reports an issue).

PR #89, opened 2019-10-26
Fix JavaDoc warnings/errors causing build failures with JDK8 and JDK11

JCgH4164838Gh792C124B5:
Fix all JavaDoc warnings/errors that caused build failures when executing the Maven build with JDK8 and JDK11.
- Modified any JavaDoc entries as required in all source files (often adding missing @param, @return and @throws elements).  Some HTML escaping was also required, as well as bad tag formatting.
- Added source specifier and API link location to the project pom's javadoc-plugin entry
(see https://issues.apache.org/jira/browse/MJAVADOC-562 and https://issues.apache.org/jira/browse/MJAVADOC-615 for background).
- Build succeeded with JDK 7u79, 8u201, 11.0 and 11.0.4.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -153,6 +153,10 @@
                         <compress>true</compress>
                         <index>true</index>
                     </archive>
+                    <source>1.6</source>
+                    <links>
+                        <link>http://docs.oracle.com/javase/6/docs/api/</link>
+                    </links>
                 </configuration>
                 <executions>
                     <execution>
--- a/src/java/ognl/ASTConst.java
+++ b/src/java/ognl/ASTConst.java
@@ -56,7 +56,11 @@ public class ASTConst extends SimpleNode implements NodeType
         super(p, id);
     }
 
-    /** Called from parser actions. */
+    /** 
+     * Called from parser actions.
+     *
+     * @param value the Object representing the value.
+     */
     public void setValue(Object value)
     {
         this.value = value;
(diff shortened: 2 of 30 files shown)


JCgH4164838Gh792C124B5:
Hello OGNL Team.

The JavaDoc related OGNL build failures when trying to build with newer JDKs came up in a recent JIRA comment for Struts.

It took a little more work than expected, but this PR aims to eliminate the Maven build failures that previously occurred when building with newer JDKs (8 and 11 were the notable ones, as LTS versions of Java).

**The PR should only have modified the JavaDoc tag elements** (or whitespace related to the overall entry), so **no code logic changes should be present**.  If anyone sees and inadvertent change, please identify it in a review of the PR.

Take a look and let me know if you want any changes to the PR.  If it is accepted then hopefully an equivalent (possibly via cherry-picking) can be applied to OGNL 3.2.x.

pom.xml
```diff
@@ -153,6 +153,10 @@
                         <compress>true</compress>
                         <index>true</index>
                     </archive>
+                    <source>1.6</source>
```
lukaszlenart: I think we can switch to 1.7, it won't be possible to build and publish using 1.6 anymore

JCgH4164838Gh792C124B5: Hello @lukaszlenart .

I think you are right - it's probably OK to switch source/target to 1.7 now (it only changed a little while ago from 1.5 to 1.6 due the Travis CI builds being blocked).   But maybe it would be cleaner to do that in a separate PR that only changes the POM ?

Let me know if you prefer source/target changed to 1.7 in this PR, or wait for a separate PR afterwards.

lukaszlenart: Good point

lukaszlenart:
LGTM 👍

PR #95, opened 2019-12-27
Update source/target level for OGNL 3.1.x to JDK 1.7.

JCgH4164838Gh792C124B5:
Update source/target level for OGNL 3.1.x to JDK 1.7.
- Update pom.xml with source/target 1.7 as per L. Lenart comment in PR#89.
- Update surefire plugin to latest 2.x available.
- Update javadoc comment in Ognl class causing build failure in JDK 8/11.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -78,12 +78,12 @@
 
         <plugins>
             <plugin>
-                <!-- Raised source/target levels to 1.6 as Travis CI now rejects 1.5 -->
+                <!-- Raised source/target levels to 1.7 as Travis CI now rejects 1.5 -->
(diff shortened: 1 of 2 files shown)


JCgH4164838Gh792C124B5:
This PR is a configuration/JavaDoc comment only change that does not impact any code.
The intention is to upgrade the source/target levels to 1.7 as per L. Lenart's comment in PR #89. 

It also updates surefire to the latest 2.x and fixes a JavaDoc comment causing failures when building with Oracle JDK8.  The build was tested locally with JDK 7, 8 and 11.

lukaszlenart:
LGTM 👍

Issue #81, opened 2019-09-24
OGNLContext deprecated constructor crashes


deved65:
In current release 3.2.11 and master the deprecated constructor  `public OgnlContext(Map values)` in line 117 is unusable and should be removed immediately (or fixed). Currently it will always leads to a NullPointerException in line 170 of the class.  To my understanding functionality marked with 'deprecated' should no longer be used (as it will be removed in a future release), but should not be unusable 'by concept'.

lukaszlenart:
Are you sure it's about the [line 170](https://github.com/jkuhnert/ognl/blob/master/src/main/java/ognl/OgnlContext.java#L170)?

deved65:
Sorry, I messed it up. It is [line 140](https://github.com/jkuhnert/ognl/blob/888b193f951c854f1acfd0d236bd1592f9632a17/src/main/java/ognl/OgnlContext.java#L120) with a RuntimeException. In [line 120 ](https://github.com/jkuhnert/ognl/blob/888b193f951c854f1acfd0d236bd1592f9632a17/src/main/java/ognl/OgnlContext.java#L140) the extensive [constructor](https://github.com/jkuhnert/ognl/blob/888b193f951c854f1acfd0d236bd1592f9632a17/src/main/java/ognl/OgnlContext.java#L123) is invoked with null reference for parameter `memberAccess`. In line 137 is a null reference check on memberAccess which in turn always must lead to the RuntimeException.

lukaszlenart:
The best option I see is to drop that constructor..

deved65:
If no meaningful or 'cheap' default implementation for MemberAccess can be provided I tend to agree.

lukaszlenart:
Yeah... that's why I have removed the default implementation from the main code and moved it into tests - it was quite often used to replace a framework's specific implementation by hackers. And that's why OGNL 3.2.x requires to provide your own implementation, even as a copy of the default one based on the one used in tests.

lukaszlenart:
Also please be aware that 3.2.x line is still a [work-in-progress](https://github.com/jkuhnert/ognl/blob/master/docs/VersionNotes.md#release-notes---version-323---wip-new-changes-are-coming) - maybe I should express that better

JCgH4164838Gh792C124B5:
Hi.

[...]

PR #101, opened 2020-02-23
Proposed fix for Issue #81 (deprecated constructor always throws an exception)

JCgH4164838Gh792C124B5:
Proposed fix for Issue #81:
- Drops OgnlContext(Map values) constructor that always fails with an
Exception.
- Additional comments for remaining two constructors.
- Changed constructor to ensure _values Map can no longer be null.
- Added comments to setValues(Map value) and getValues() for clarity.
- Added @Override annotations to Map interface methods.

### DIFF
--- a/src/main/java/ognl/OgnlContext.java
+++ b/src/main/java/ognl/OgnlContext.java
@@ -102,11 +102,11 @@ public class OgnlContext extends Object implements Map
 
     /**
      * Constructs a new OgnlContext with the given class resolver, type converter and member access.
-     * If any of these parameters is null the default will be used.
+     * If any of these parameters is null the default will be used, except <span class="strong">memberAccess which must be non-null</span>.
      * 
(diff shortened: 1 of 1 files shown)


JCgH4164838Gh792C124B5:
Hello OGNL Team.

This PR should address the concern raised in the issue with a little low-risk cleanup added as well.

Other than dropping the deprecated constructor, this PR also ensures the context's values `Map` cannot be initialized to `null` (given the usage in the rest of the class, allowing a `null` values `Map` does not seem to make sense).  That _could_ impact someone who intentionally set the values to `null`, but preventing this in 3.2.x seems worth that small risk (and the JavaDoc comments indicate `null` parameters will be set to defaults - except for `MemberAccess` which cannot be `null`).

Let me know what you think.

lukaszlenart:
Struts build passed with a SNAPSHOT version, I will wait a bit before pushing a new release

PR #106, opened 2020-09-30
resolve race condition problem

rolandhe:
I found a race condition problem which cause to throw NoSuchPropertyException when there are to many threads since version 3.1.2. Although it was resolved by trying to call method of "getReadMethod", it will loss some performance,  I think it is the best way  to resolve  race condition.

### DIFF
--- a/src/main/java/ognl/OgnlRuntime.java
+++ b/src/main/java/ognl/OgnlRuntime.java
@@ -2767,12 +2767,15 @@ public class OgnlRuntime {
     {
         // Cache is a map in two levels, so we provide two keys (see comments in ClassPropertyMethodCache below)
         Method method = cacheGetMethod.get(targetClass, propertyName);
+        if (method == ClassPropertyMethodCache.NULL_REPLACEMENT) {
+            return null;
+        }
         if (method != null)
             return method;
 
         // By checking key existence now and not before calling 'get', we will save a map resolution 90% of the times
-        if (cacheGetMethod.containsKey(targetClass, propertyName))
-            return null;
+//        if (cacheGetMethod.containsKey(targetClass, propertyName))
+//            return null;
 
         method = _getGetMethod(context, targetClass, propertyName); // will be null if not found - will cache it anyway
         cacheGetMethod.put(targetClass, propertyName, method);
@@ -2868,12 +2871,15 @@ public class OgnlRuntime {
     {
         // Cache is a map in two levels, so we provide two keys (see comments in ClassPropertyMethodCache below)
         Method method = cacheSetMethod.get(targetClass, propertyName);
+        if (method == ClassPropertyMethodCache.NULL_REPLACEMENT) {
+            return null;
+        }
         if (method != null)
             return method;
 
         // By checking key existence now and not before calling 'get', we will save a map resolution 90% of the times
-        if (cacheSetMethod.containsKey(targetClass, propertyName))
-            return null;
+//        if (cacheSetMethod.containsKey(targetClass, propertyName))
+//            return null;
 
         method = _getSetMethod(context, targetClass, propertyName); // will be null if not found - will cache it anyway
         cacheSetMethod.put(targetClass, propertyName, method);
@@ -3948,12 +3954,9 @@ public class OgnlRuntime {
             ConcurrentHashMap<String,Method> methodsByPropertyName = this.cache.get(clazz);
             if (methodsByPropertyName == null)
             {
-                methodsByPropertyName = new ConcurrentHashMap<String, Method>();
-                this.cache.put(clazz, methodsByPropertyName);
+                return null;
             }
             Method method = methodsByPropertyName.get(propertyName);
-            if (method == NULL_REPLACEMENT)
-                return null;
             return method;
         }
 
@@ -3963,19 +3966,22 @@ public class OgnlRuntime {
             if (methodsByPropertyName == null)
             {
                 methodsByPropertyName = new ConcurrentHashMap<String, Method>();
-                this.cache.put(clazz, methodsByPropertyName);
+                ConcurrentHashMap<String,Method> old = this.cache.putIfAbsent(clazz, methodsByPropertyName);
+                if (null != old) {
+                    methodsByPropertyName = old;
+                }
             }
-            methodsByPropertyName.put(propertyName, (method == null? NULL_REPLACEMENT : method));
+            methodsByPropertyName.putIfAbsent(propertyName, (method == null? NULL_REPLACEMENT : method));
         }
 
-        boolean containsKey(Class clazz, String propertyName)
-        {
-            ConcurrentHashMap<String,Method> methodsByPropertyName = this.cache.get(clazz);
-            if (methodsByPropertyName == null)
-                return false;
-
-            return methodsByPropertyName.containsKey(propertyName);
-        }
+//        boolean containsKey(Class clazz, String propertyName)
+//        {
+//            ConcurrentHashMap<String,Method> methodsByPropertyName = this.cache.get(clazz);
+//            if (methodsByPropertyName == null)
+//                return false;
+//
+//            return methodsByPropertyName.containsKey(propertyName);
+//        }
 
         /**
          * Allow clearing for the underlying cache of the ClassPropertyMethodCache.
--- a/src/main/java/ognl/OgnlRuntime.java
+++ b/src/main/java/ognl/OgnlRuntime.java
@@ -2773,10 +2773,6 @@ public class OgnlRuntime {
         if (method != null)
             return method;
 
-        // By checking key existence now and not before calling 'get', we will save a map resolution 90% of the times
-//        if (cacheGetMethod.containsKey(targetClass, propertyName))
-//            return null;
-
         method = _getGetMethod(context, targetClass, propertyName); // will be null if not found - will cache it anyway
         cacheGetMethod.put(targetClass, propertyName, method);
 
@@ -3974,14 +3970,6 @@ public class OgnlRuntime {
             methodsByPropertyName.putIfAbsent(propertyName, (method == null? NULL_REPLACEMENT : method));
         }
 
-//        boolean containsKey(Class clazz, String propertyName)
-//        {
-//            ConcurrentHashMap<String,Method> methodsByPropertyName = this.cache.get(clazz);
-//            if (methodsByPropertyName == null)
-//                return false;
-//
-//            return methodsByPropertyName.containsKey(propertyName);
-//        }
 
         /**
          * Allow clearing for the underlying cache of the ClassPropertyMethodCache.
--- a/src/test/java/ognl/race/Base.java
+++ b/src/test/java/ognl/race/Base.java
@@ -1,0 +1,15 @@
+package ognl.race;
+
+/**
+ *
+ */
+public class Base {
+    private Boolean yn = true;
+    public Boolean getYn() {
+        return yn;
+    }
+
+    public void setYn(Boolean yn) {
+        this.yn = yn;
+    }
+}
--- a/src/test/java/ognl/race/Persion.java
+++ b/src/test/java/ognl/race/Persion.java
@@ -1,0 +1,14 @@
+package ognl.race;
+
+public class Persion extends Base{
+    private String name = "abc";
+
+
+    public String getName() {
+        return name;
+    }
+
+    public void setName(String name) {
+        this.name = name;
+    }
+}
--- a/src/test/java/ognl/race/RaceTestCase.java
+++ b/src/test/java/ognl/race/RaceTestCase.java
@@ -1,0 +1,74 @@
+package ognl.race;
+
+
+import ognl.DefaultMemberAccess;
+import ognl.Ognl;
+import ognl.OgnlContext;
+import ognl.OgnlException;
+import org.junit.Assert;
+import org.junit.Test;
+
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.atomic.AtomicInteger;
+
+
+public class RaceTestCase {
+
+    @Test
+    public void testOgnlRace(){
+        int concurrent = 128;
+        final int batchCount = 2000;
+        final CountDownLatch start = new CountDownLatch(1);
+        final CountDownLatch wait = new CountDownLatch(concurrent);
+        final AtomicInteger errCount = new AtomicInteger(0);
+
+        final Persion persion = new Persion();
+        for (int i = 0; i < concurrent;i++){
+            Thread thread = new Thread(new Runnable() {
+                @Override
+                public void run() {
+                    try {
+                        start.await();
+                    } catch (InterruptedException e) {
+                        e.printStackTrace();
+                    }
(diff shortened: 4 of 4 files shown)


src/main/java/ognl/OgnlRuntime.java
```diff
@@ -2767,12 +2767,15 @@ public static Method getGetMethod(OgnlContext context, Class targetClass, String
...
+            return null;
+        }
         if (method != null)
             return method;
 
         // By checking key existence now and not before calling 'get', we will save a map resolution 90% of the times
-        if (cacheGetMethod.containsKey(targetClass, propertyName))
-            return null;
+//        if (cacheGetMethod.containsKey(targetClass, propertyName))
+//            return null;
```
lukaszlenart, line 2778: Please remove these line instead of commenting them out. Also remove the comment above.

src/main/java/ognl/OgnlRuntime.java
```diff
@@ -2868,12 +2871,15 @@ public static Method getSetMethod(OgnlContext context, Class targetClass, String
...
+            return null;
+        }
         if (method != null)
             return method;
 
         // By checking key existence now and not before calling 'get', we will save a map resolution 90% of the times
-        if (cacheSetMethod.containsKey(targetClass, propertyName))
-            return null;
+//        if (cacheSetMethod.containsKey(targetClass, propertyName))
+//            return null;
```
lukaszlenart, line 2882: Please remove these line instead of commenting them out. Also remove the comment above.

src/main/java/ognl/OgnlRuntime.java
```diff
@@ -3948,12 +3954,9 @@ Method get(Class clazz, String propertyName)
...
             if (methodsByPropertyName == null)
             {
-                methodsByPropertyName = new ConcurrentHashMap<String, Method>();
-                this.cache.put(clazz, methodsByPropertyName);
+                return null;
             }
             Method method = methodsByPropertyName.get(propertyName);
-            if (method == NULL_REPLACEMENT)
-                return null;
             return method;
```
lukaszlenart, line 3960: This breaks the logic, you will return `NULL_REPLACEMENT` instead of `null`

src/main/java/ognl/OgnlRuntime.java
```diff
@@ -3963,19 +3966,22 @@ void put(Class clazz, String propertyName, Method method)
...
-            return methodsByPropertyName.containsKey(propertyName);
-        }
+//        boolean containsKey(Class clazz, String propertyName)
+//        {
+//            ConcurrentHashMap<String,Method> methodsByPropertyName = this.cache.get(clazz);
+//            if (methodsByPropertyName == null)
+//                return false;
+//
+//            return methodsByPropertyName.containsKey(propertyName);
+//        }
```
lukaszlenart, line 3984: Please remove the code instead of commenting it out

lukaszlenart (changes requested):

lukaszlenart:
@harawata @JCgH4164838Gh792C124B5 @yasserzamani maybe you would like to also check this change.

harawata:
Thank you, @rolandhe !
And thank you for the heads-up, @lukaszlenart !

I just took a quick look and didn't find any logical issue. 👍

src/main/java/ognl/OgnlRuntime.java
yasserzamani: :+1:

JCgH4164838Gh792C124B5:
Hello @rolandhe .  Thank you for reporting the race condition scenario (and thanks to @lukaszlenart for advising about the PR conversation).  👍 

I read the proposed code changes in the PR and subsequent comments, but remain unsure if the proposed changes will work as intended.  Can you tell us approximately how many concurrent threads were involved (10's, 100's, 1000's, etc.) before the symptom started being evident ?

Also, _could you provide a small example_ (inline in a comment here, or maybe a link to small reproducer project) _that reproduces the condition_ (causing `NoSuchPropertyException` to be thrown with some consistency) ?

Since _changes to the `getGetMethod()` and `getSetMethod()` could have some subtle side-effects to the behaviour and performance of OGNL_ (and the applications that use it), some extra caution may be needed when evaluating the proposed changes.  A reproducer, that can be used to cause the condition, would help evaluate proposed changes.

There may be a trade-off involved in changes to the "two-level-cache logic" that is currently in place for OGNL.  Structural changes to avoid the `NoSuchPropertyException` condition could change the performance characteristics of OGNL.  For that reason, the probability/frequency of the issue in OGNL 3.1.x should probably be considered against performance impacts.

_If the race condition can be eliminated (or reduced) without any substantial reduction in performance, or other undesirable side-effects, that would be ideal_.  If not, then it may be a "judgment call" as to what is "better" (and maybe even something requiring a flag/choice).

Thanks again, @rolandhe.  Any additional details you can provide will help out.  😄

src/main/java/ognl/OgnlRuntime.java
```diff
@@ -3963,19 +3966,22 @@ void put(Class clazz, String propertyName, Method method)
             if (methodsByPropertyName == null)
             {
                 methodsByPropertyName = new ConcurrentHashMap<String, Method>();
-                this.cache.put(clazz, methodsByPropertyName);
+                ConcurrentHashMap<String,Method> old = this.cache.putIfAbsent(clazz, methodsByPropertyName);
```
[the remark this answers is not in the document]
rolandhe: `/**
     * {@inheritDoc}
     *
     * @return the previous value associated with the specified key,
     *         or {@code null} if there was no mapping for the key
     * @throws NullPointerException if the specified key or value is null
     */
    public V putIfAbsent(K key, V value) `

"if" check and then put, this is classic race condition.

src/main/java/ognl/OgnlRuntime.java
rolandhe: done

src/main/java/ognl/OgnlRuntime.java
rolandhe: done

src/main/java/ognl/OgnlRuntime.java
rolandhe: done

rolandhe:
> Hello @rolandhe . Thank you for reporting the race condition scenario (and thanks to @lukaszlenart for advising about the PR conversation). 👍
> 
> I read the proposed code changes in the PR and subsequent comments, but remain unsure if the proposed changes will work as intended. Can you tell us approximately how many concurrent threads were involved (10's, 100's, 1000's, etc.) before the symptom started being evident ?
> 
> Also, _could you provide a small example_ (inline in a comment here, or maybe a link to small reproducer project) _that reproduces the condition_ (causing `NoSuchPropertyException` to be thrown with some consistency) ?
> 
> Since _changes to the `getGetMethod()` and `getSetMethod()` could have some subtle side-effects to the behaviour and performance of OGNL_ (and the applications that use it), some extra caution may be needed when evaluating the proposed changes. A reproducer, that can be used to cause the condition, would help evaluate proposed changes.
> 
> There may be a trade-off involved in changes to the "two-level-cache logic" that is currently in place for OGNL. Structural changes to avoid the `NoSuchPropertyException` condition could change the performance characteristics of OGNL. For that reason, the probability/frequency of the issue in OGNL 3.1.x should probably be considered against performance impacts.
> 
> _If the race condition can be eliminated (or reduced) without any substantial reduction in performance, or other undesirable side-effects, that would be ideal_. If not, then it may be a "judgment call" as to what is "better" (and maybe even something requiring a flag/choice).
> 
> Thanks again, @rolandhe. Any additional details you can provide will help out. 😄

I committed example that reproduce condition exception, it is ognl.race.RaceTestCase#testOgnlRace,,before you run this code you should remove lines 2147 and 2148 of OgnlRuntime.java. Maybe you would run many times and then produce the "NoSuchPropertyException".

src/main/java/ognl/OgnlRuntime.java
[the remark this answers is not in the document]
yasserzamani: No I meant isn't following piece redundant:
```java
            ConcurrentHashMap<String,Method> methodsByPropertyName = this.cache.get(clazz);
            if (methodsByPropertyName == null)
            {
```
because you've already used `putIfAbsent` inside the `if`.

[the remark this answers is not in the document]
rolandhe: Using read-compare-write style, which would reduce “new” operation, and very few method call new concurrent
map object.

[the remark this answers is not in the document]
yasserzamani: :clap: thanks!

JCgH4164838Gh792C124B5:
Hello @rolandhe (and everyone).

Thanks for providing the reproducer test code.  😄 

Unfortunately, using the reproducer code I was unable to reproduce any failures or exceptions with the existing OGNL 3.2.16-SNAPSHOT code (many executions, with variations from a few threads to 1000s).  Adding an additional test that called the original test multiple times with varied numbers of threads and iterations, along with cache clears, did not generate any exceptions either.

I also tried the test code with the `OgnlRuntime` from this PR (both "as-is" and with lines 2147-2148 commented out), using the same sorts of tests against the original code.  There were no failures or exceptions in that case either.

It seems that in practical terms, the probability of the race condition in the original code appears to be very slight, otherwise we would expect to be able to reproduce it with some frequency (at least with 100's or 1000's of threads and many repeated runs).

_I am not sure if the reproducer test, given it cannot (consistently) demonstrate the presence or absence of the race condition failure, should be included in the final revision of this PR_ ?   Others might have different thoughts on that.

The good news is that after running multiple builds, with both the old code and the new code from this PR, there were no indications of any observable performance differences between the two.   😄 

With no apparent performance impact seen, the proposed `OgnlRuntime` logic change seems OK to me.  Good work!   👍  👏

harawata:
Just FYI, here is the repro I used when investigating the `NoSuchPropertyException` before [(1)](https://github.com/mybatis/mybatis-3/issues/623) [(2)](https://github.com/mybatis/mybatis-3/issues/1648) .

```java
import ognl.*;
import java.util.*;
import java.util.concurrent.*;
import java.util.stream.*;
import java.lang.reflect.*;

public class GetGetMethodRaceCondition {
  public static void main(String[] args) throws Exception {
    int run = Integer.valueOf(args[0]);
    List<Future<Object>> futures = new ArrayList<>();
    ExecutorService executor = Executors.newCachedThreadPool();
    try {
      IntStream.range(0, run).forEach(i -> {
        futures.add(executor.submit(() -> {
          return OgnlRuntime.getGetMethod(null, String.class, "bytes");
        }));
      });
      for (int i = 0; i < run; i++) {
        if (futures.get(i).get() == null) {
          System.out.println("!!! ERROR !!!! at round " + i);
        }
      }
    } finally {
      executor.shutdown();
    }
  }
}
```

To reproduce the issue reliably, I had to use a shell script like the below.

```sh
#!/bin/sh

# compile
javac GetGetMethodRaceCondition.java -cp ognl-3.2.15.jar

echo ''
# run rounds
for i in {1..100}
do
  java -cp ognl-3.2.15.jar:javassist-3.27.0-GA.jar:. GetGetMethodRaceCondition 100
done
```

p.s.
The direct cause of `NoSuchPropertyException` was in `getReadMethod` and it was fixed in #104 , so 3.2.15 won't throw `NoSuchProprtyException` even with this race condition exist, I believe.

rolandhe:
> Hello @rolandhe (and everyone).
> 
> Thanks for providing the reproducer test code. 😄
> 
> Unfortunately, using the reproducer code I was unable to reproduce any failures or exceptions with the existing OGNL 3.2.16-SNAPSHOT code (many executions, with variations from a few threads to 1000s). Adding an additional test that called the original test multiple times with varied numbers of threads and iterations, along with cache clears, did not generate any exceptions either.
> 
> I also tried the test code with the `OgnlRuntime` from this PR (both "as-is" and with lines 2147-2148 commented out), using the same sorts of tests against the original code. There were no failures or exceptions in that case either.
> 
> It seems that in practical terms, the probability of the race condition in the original code appears to be very slight, otherwise we would expect to be able to reproduce it with some frequency (at least with 100's or 1000's of threads and many repeated runs).
> 
> _I am not sure if the reproducer test, given it cannot (consistently) demonstrate the presence or absence of the race condition failure, should be included in the final revision of this PR_ ? Others might have different thoughts on that.
> 
> The good news is that after running multiple builds, with both the old code and the new code from this PR, there were no indications of any observable performance differences between the two. 😄
> 
> With no apparent performance impact seen, the proposed `OgnlRuntime` logic change seems OK to me. Good work! 👍 👏

Please copy test code to your MASTER branch code, NOT this PR code, then remove or comment next two lines:

`
   if (m == null)
            m = getReadMethod((target == null) ? null : target.getClass(), propertyName, null);
`

then run the test code, maybe run many times, I tested five times, and reproduce.

JCgH4164838Gh792C124B5:
Hi @rolandhe .

Sorry if it was not clear from the earlier comment, but I previously had copied and ran the test code against the master branch of OGNL 3.2.  I tested both with the original branch state for `getMethodValue()`, as well as with the two lines commented out, but did not see any failures in multiple runs.

Today I repeated the same tests again many times manually, and was finally able to reproduce the failure (using the original OGNL 3.2 OgnlRuntime, with the two lines commented out of `getMethodValue()`).  Without commenting out those two lines, I never saw the error being produced.

I also performed the test with the OgnlRuntime code from this PR in its original form, and with the same lines commented out in `getMethodValue()` for comparison, and did not see any failures in multiple runs.

Using a process like the sample @harawata provided above (thanks @harawata 😄 ), it was easier to produce the error more consistently as well.

Anyway, your PR's changes appears to eliminate the race without any noticeable side-effects.  Thanks for your work on this.  👏

src/main/java/ognl/OgnlRuntime.java
lukaszlenart: What about this? As far I understand this will return `NULL_REPLACEMENT` which was set in [`put`](https://github.com/jkuhnert/ognl/pull/106/files#diff-9464615a5f9bf81dc8354e523e412fa747d62b868f3fbb3270d8a2dd0bda9ee9R3970)

rolandhe: Sorry,I can't catch what you said.
Yes, It will return NULL_REPLACEMENT if you put.

lukaszlenart: As far I understand, the previous code will return a raw `null` if `method` from cache equals to `NULL_REPLACEMENT` (which is initialised to `ClassPropertyMethodCache#get`). You've changed the code to return `method` directly without checking if this is `NULL_REPLACEMENT`, so this means you will return `NULL_REPLACEMENT` instead of `null`.

harawata: It's now checked in the callers (getGetMethod and getSetMethod).

lukaszlenart: Thanks @harawata for clarification :)

lukaszlenart (approved):
LGTM 👍

PR #107, opened 2020-10-12
Minor cleanup for 3.2.x, related to previous 3.1.x merges

JCgH4164838Gh792C124B5:
Minor cleanup fix:
- Removal of unneeded call (todo) in OgnlRuntime.invokeMethod() since the
SecurityManager sandbox logic has been merged from 3.1.x to 3.2.x.  Same
logic as in 3.1.x.
- Avoid extra getModifiers() call in OgnlRuntime.setFieldValue(), same
logic as in 3.1.x.


lukaszlenart (approved):
LGTM 👍

PR #109, opened 2020-10-22
Introduces AbstractMemberAccess to allow create it on-fly

### LINKED ISSUE #105: IllegalArgumentException when calling Ognl.getValue(Object, Object, Class)
Tested with 3.2.14.

Calling `Ognl.getValue(Object, Object, Class)` calls `Ognl.createDefaultContext(Object)` (which is marked as deprecated). This then calls constructor `OgnlContext(ClassResolver, TypeConverter, MemberAccess)` with all `null` arguments, which finally calls the other constructor `OgnlContext(MemberAccess, ClassResolver, TypeConverter, Map)` which throws `IllegalArgumentException("MemberAccess implementation must be provided - null not permitted!")`.

I'm not sure whether the internal use of a deprecated call is a bug (deprecated is not supposed to mean _broken_, after all), or whether further overloads of `Ognl.getValue` (I'm actually calling `Ognl.getValue(String, Object)` which calls `Ognl.getValue(String, Object, Class)`, which calls `Ognl.getValue(Object, Object, Class)`) ought to have been deprecated (although ditto), or whether something else is amiss, but there is clearly an issue here.

The issue was introduced in version 3.2.2, in commit e30b2112979d1be7bed8073f1f47bffed6277fb2

lukaszlenart:
Closes #105

### DIFF
--- a/src/main/java/ognl/AbstractMemberAccess.java
+++ b/src/main/java/ognl/AbstractMemberAccess.java
@@ -1,0 +1,36 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * and/or LICENSE file distributed with this work for additional
+ * information regarding copyright ownership.  The ASF licenses
+ * this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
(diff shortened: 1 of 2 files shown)


lukaszlenart (commented):
No objections, LGTM 👍

Issue #117, opened 2021-03-17
The 3.2.18 version of ognl.Ognl#getValue(..) is not easy to use at all 


zhuster:
It may need to be redesigned

lukaszlenart:
Could you share some more details?

zhuster:
I have created [a pull request](https://github.com/jkuhnert/ognl/pull/118).

lukaszlenart:
Thanks, merged!

PR #118, opened 2021-03-18
Fix createDefaultContext, otherwise it will always throw an exception

zhuster:


lukaszlenart (approved):
LGTM 👍 
Closes #117

lukaszlenart:
New release is under way
https://github.com/jkuhnert/ognl/blob/master/docs/VersionNotes.md#release-notes---version-3219-2021-03-18

PR #119, opened 2021-03-23
Change to ensure two existing tests are called by automation, plus minor test correction

JCgH4164838Gh792C124B5:
Existing test processing issue fixes:
- Add two tests that were missing from surefire builds.
- Fix test_ApplyExpressionMaxLength() test issue not seen due to test not running during automated builds.
- Add finally guard to Improve test_ApplyExpressionMaxLength() and test_FreezeThawExpressionMaxLength() to restore default state even if the test fails, otherwise it can make other tests appear to fail due to side-effects.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -112,6 +112,8 @@
                     <includes>
                         <include>**/*Test.java</include>
                         <include>**/TestOgnlRuntime.java</include>
+                        <include>**/TestOgnlException.java</include>
+                        <include>**/TestExpressionCompiler.java</include>
                     </includes>
                 </configuration>
             </plugin>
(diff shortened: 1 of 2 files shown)


JCgH4164838Gh792C124B5:
Hello OGNL Team.  While trying something locally, I discovered that `TestExpressionCompiler` was not being executed during the standard build (and after further review that `TestOgnlException` was also not being executed, either).

This PR seeks to correct that as well as an issue with one test in `TestExpressionCompiler` found at the same time (not noted previously due to the tests not being run during the build).

If the PR is accepted, the equivalent changes should be cherry-picked into the 3.2.x branch, since its build behaves the same way as well.

lukaszlenart:
Bravo 💯  LGTM 👍

PR #122, opened 2021-04-02
LanguageGuide.md: Backslash escape vertical bars in pipe table

jessepav:
Even though they were within backtick code spans, the '|' characters were interpreter as table delimeters.


lukaszlenart (approved):
Good catch 👏 
LGTM 👍

Issue #103, opened 2020-05-21
Class reference is failed when include 'or' at a package name


kazuki43zoo:
If package name includes "or" (e.g. "jp.or.example"), the `Ognl#parseExpression(String)` is failed as follow:

* Code

```java
    public void test_class_reference()
        throws Throwable
    {
        OgnlContext context = (OgnlContext) Ognl.createDefaultContext(null, new DefaultMemberAccess(false));
        {
            Node expr = (Node) Ognl.parseExpression("@java.util.UUID@randomUUID()"); // It's OK
            Object v = expr.getValue(context, null);
            System.out.println(v);
        }
        {
            Node expr = (Node) Ognl.parseExpression("@jp.or.example.IdUtils@generateId()"); // It's failed on parsing phase
            Object v = expr.getValue(context, null);
            System.out.println(v);
        }
    }
```

* Stack Trace

```
ognl.ExpressionSyntaxException: Malformed OGNL expression: @jp.or.example.IdUtils@generateId()
 [ognl.ParseException: Encountered " "or" "or "" at line 1, column 5.
Was expecting:
    <IDENT> ...
    ]
	at ognl.Ognl.parseExpression(Ognl.java:179)
	at org.ognl.test.enhance.TestExpressionCompiler.test_class_reference(TestExpressionCompiler.java:55)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at junit.framework.TestCase.runTest(TestCase.java:176)
	at junit.framework.TestCase.runBare(TestCase.java:141)
	at junit.framework.TestResult$1.protect(TestResult.java:122)
	at junit.framework.TestResult.runProtected(TestResult.java:142)
	at junit.framework.TestResult.run(TestResult.java:125)
	at junit.framework.TestCase.run(TestCase.java:129)
	at junit.framework.TestSuite.runTest(TestSuite.java:252)
	at junit.framework.TestSuite.run(TestSuite.java:247)
	at org.junit.internal.runners.JUnit38ClassRunner.run(JUnit38ClassRunner.java:86)
[...]

PR #124, opened 2021-04-11
Potential mitigation for issue #103 (parse exception for package name including "or")

JCgH4164838Gh792C124B5:
Potential mitigation for issue #103 (identifiers cannot include package names like "and" or "or"):
- Sub-optimal solution requires direct modification of OgnlParser, so if the language specification (ognl.jj, ognl.jjt) ever changes and is rebuilt using JavaCC/JJTree, the logic would have to be manually re-added.
- Mitigation only works for sub-package names, still fails for patterns like "or.something".


JCgH4164838Gh792C124B5:
Hello OGNL Team.

This PR may not be desirable, given it has to change `OgnlParser` directly, and there may be unanticipated side-effects from the change as well.  It should be reviewed thoroughly before considering using it.

The changes pass the unit tests, and passed a Struts 2.5.27-SNAPSHOT regression build using an OGNL snapshot built from the PR code, but further testing and consideration may be needed (for example, something with many complex OGNL expressions).

If someone familiar with JavaCC/JJTree could review what the `OgnlParser` changes are attempting to do, and propose a way to change the language specification files (`ognl.jj`, `ognl.jjt`) to achieve the same (or better) result, that would be great.

lukaszlenart:
Thanks a lot for this PR, yet I have a question about `or.package` or `and.package` - what's the main issue to recognise such expressions? Isn't just a matter of checking if `.` follows the package name?

JCgH4164838Gh792C124B5:
Hi @lukaszlenart .

The main issue appears to be that when the OGNL language parser encounters the patterns that match a token name (of which "and" or "or" are only two of many), it interprets them as tokens, even if they are part of a package name.

My attempts to figure out how the JavaCC/JJTree language specification might be changed to differentiate between package "xxx.and.yyy and an expression "x and y" fell short, so I resorted to tracing breakpoints in the generated code to see if the state machine flow could be "safely altered" to have the desired result. 

I still dislike the solution I came up with, since it deviates from the JavaCC/JJTree language specification.  I think it is too risky to actually use this PR's solution, but was hoping a JavaCC/JJTree expert might see it, and be able to propose a proper fix.

Without a proper fix / solution, it might be better to simply document the limitation that OGNL cannot process any package names whose components exactly match OGNL token names.

lukaszlenart:
I'm not a JavaCC/JJTree expert, I would like to understand how does it work but this is really out of my skills :) That being said and taking your doubts into account I would just left this PR open hoping someone will come up with a better solution and document OGNL limitations.

lukaszlenart:
I added a note about those limitations https://github.com/jkuhnert/ognl/pull/135

PR #139, opened 2021-11-19
Explicitly cast the Enum#compareTo argument

harawata:
Hello @lukaszlenart ,

javac works fine, but Eclipse JDT requires explicit casting.
It seems that the JLS is not clear about this.


lukaszlenart:
LGTM 👍

lukaszlenart:
OGNL 3.3.0 is out and it also requires Java 8
https://github.com/jkuhnert/ognl/blob/master/docs/VersionNotes.md#release-notes---version-330-2021-11-19

PR #140, opened 2021-11-23
[OGNL-102] Improves performance when null was returned

lukaszlenart:
Improves performance of chained expressions in case if the `root` is `null`

Refs [OGNL-102](https://issues.apache.org/jira/browse/OGNL-102)

### DIFF
--- a/src/main/java/ognl/ASTChain.java
+++ b/src/main/java/ognl/ASTChain.java
@@ -78,10 +78,15 @@ public class ASTChain extends SimpleNode implements NodeType, OrderedReturn
     protected Object getValueBody(OgnlContext context, Object source)
             throws OgnlException
     {
-        Object result = source;
+        // short-circuit the chain only in case if the root is null
+        if (source == null) {
+            return null;
+        }
 
+        Object result = source;
         for(int i = 0, ilast = _children.length - 1; i <= ilast; ++i)
         {
+
             boolean handled = false;
 
             if (i < ilast) {
@@ -124,7 +129,7 @@ public class ASTChain extends SimpleNode implements NodeType, OrderedReturn
                                             + "'"); }
                                 }
                             }
-                            if (!handled) 
+                            if (!handled)
                             {
                                 result = OgnlRuntime.getIndexedProperty(context, result,
                                                                         propertyNode.getProperty(context, result).toString(),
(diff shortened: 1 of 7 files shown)


lukaszlenart:
@davoustp @harawata @quaff could you take a look? this should improve performance a bit, it's in relation to https://github.com/jkuhnert/ognl/pull/138 and the JIRA ticket.

harawata:
Hi @lukaszlenart ,

I'm sorry about the slow response (as always)!

So, is `root` the only concern here?

With the following test, for example, the for-loop is still evaluated for the second node and the similar exception is thrown :

> ognl.OgnlException: source is null for getProperty(null, "key2")

```java
public void testNullHnadling() throws Exception {
  OgnlContext context = (OgnlContext) Ognl.createDefaultContext(null, new DefaultMemberAccess(false));
  Map<String, Object> root = new HashMap<>();
  root.put("key1", null);
  String expr = "key1.key2.key3";
  assertNull(Ognl.getValue(expr, context, root));
}
```

lukaszlenart:
Right, but if I did that on `children` level in the `for` loop a lot of tests have started to fail, especially this [one](https://github.com/jkuhnert/ognl/pull/140/files#diff-92ad2bf9f492a4e873133771bba1a612cc41d56bcb8d7651b37ca2ec689cdd0aR71-R76)

harawata:
Yes, it will change the designed behavior and @davoustp seems to be [aware](https://github.com/apache/struts/pull/504#issuecomment-968289949).

I'll just show some tests that emphasizes the difference in behavior.
If these results match @davoustp 's expectation, this PR should be fine.

```java
public void testNullValue() throws Exception {
  OgnlContext context = (OgnlContext) Ognl.createDefaultContext(null, new DefaultMemberAccess(false));
  Map<String, Object> root = new HashMap<>();
  root.put("key1", null);
  String expr = "key1.key2.key3";
  assertNull(Ognl.getValue(expr, context, root)); // FAIL : OgnlException
}

public void testEmptyRoot() throws Exception {
    OgnlContext context = (OgnlContext) Ognl.createDefaultContext(null, new DefaultMemberAccess(false));
    Map<String, Object> root = new HashMap<>();
    String expr = "key1.key2.key3";
    assertNull(Ognl.getValue(expr, context, root)); // FAIL : OgnlException
}

public void testNullRoot() throws Exception {
    OgnlContext context = (OgnlContext) Ognl.createDefaultContext(null, new DefaultMemberAccess(false));
    Map<String, Object> root = null;
    String expr = "key1.key2.key3";
    assertNull(Ognl.getValue(expr, context, root)); // PASS
}
```

lukaszlenart:
Changes will be moved into #155

PR #155, opened 2022-04-06
Merges changes from commons-ognl and cleans up code

lukaszlenart:
Commons OGNL supposed to supersede OGNL 3.x and be used instead of the 3.x branch, yet that never happened even if some preparations have been made, I took a time to merge some important fixes back into 3.x branch and fix some issues


lukaszlenart:
@davoustp it should be interested for you as well

PR #156, opened 2022-04-08
memberOf(s*)

seanhailecodes:
Thought I'd share an ognl expression we used, but I did not see here - apologies if this is the wrong spot for it


lukaszlenart:
What's the purpose of this file/change?

seanhailecodes:
Just hoping to share some code with our fellow OGNL community - this file
restricts access to specified groups

PR #144, opened 2022-01-08
Avoid illegal reflective access when possible

harawata:
As reported in #143 , making a method accessible is problematic in recent versions of Java ('illegal reflective access' warning on 9+ and InaccessibleObjectException on 16+).
This PR avoids some illegal reflective accesses that I found.
Please see the log message of each commit for the details.

PR #159, opened 2022-07-09
Fixes IllegalArgumentException : Can't decide wich method to use

harawata:
A user found another corner case [issue](https://github.com/mybatis/mybatis-3/issues/2609). 
This may be a side effect of #144 .


lukaszlenart:
LGTM 👍

lukaszlenart:
OGNL 3.3.3 is out :)

PR #193, opened 2023-01-19
Turns OgnlContext into independent class not inherited from Map

lukaszlenart:


jstenzel:
What replaces the constructor Ognl.getValue, which I used to be able to simply pass a map to, or how do I get from the simple map to the OgnlContext now?

lukaszlenart:
Create an instance of `OgnlContext` and use `setValues()` - I can always add some helper method if you want to, like `OgnlContext.of(Map)`?

jstenzel:
Helper methods are always helpful, but your info is enough for now. Many thanks

lukaszlenart:
I add a small helper #220

PR #204, opened 2023-04-11
OGNL-102 Improves performance when null was returned

lukaszlenart:
This PR introduces options to short-circuit chain operations on `null` result in case if accessing property of the object, eg. `obj1.property` where `obj1` is null or `obj1.obj2.property` where `obj2` is null.

This functionality is enabled by default, use `-Dognl.chain.short-circuit=false` to disable it.

Refs https://github.com/orphan-oss/ognl/pull/140
Refs [OGNL-102](https://issues.apache.org/jira/browse/OGNL-102)


lukaszlenart:
@davoustp @harawata @quaff could you take a look? This is another approach to improve performance, it's in relation to #138 and the JIRA ticket. @harawata I have added your test cases from #140 and they passed.

Issue #160, opened 2022-07-14
In 3.3.3 (and before) compileExpression fails without add-opens on Java 17


petergeneric:
When Ognl.compileExpression calls ExpressionCompiler.compileExpression on Java 17 with 3.3.3 which uses the latest javassist, compile fails unless add-opens args have been added to the Java commandline.

The exception arises from this line in ExpressionCompiler:

```
Class clazz = pool.toClass(newClass);
```

I think this can be avoided - Javassist has an overload for `ClassPool.toClass` that lets you define a Neighbour Class, at which point it will use the new Java MethodHandles approach to define the class rather than the legacy ClassLoader.defineClass (which requires visibility changes, which is what fails without add-opens). If I instead change that line in ExpressionCompiler to:
```
Class clazz = pool.toClass(newClass, OgnlContext.class);
```

Then the expression compiles and works perfectly under Java 17.

I don't know for sure that it's safe to use OgnlContext.class as a general purpose neighbour class (don't know if MethodHandles defineClass is different in some fundamental way to the ClassLoader defineClass, and therefore how it would interact with other  environments that make more extensive use of Modules), so could we have a way to opt-in to it, perhaps the ability to pass a neighbour class in through an Ognl.compileExpression overload (either a class or a flag that would have the internal code pass OgnlContext.class as the neighbour) -- or even some public static flag on ExpressionCompiler?

lukaszlenart:
What about extending `OgnlExpressionCompiler` with 
```java
default Optional<Class<?>> getClassNeighbor(Class<?> clazz) {
    return Optional.empty();
}
```

and the use this in
```java
Optional<Class<?>> classNeighbor = getClassNeighbor(expression.getClass());
Class<?> clazz;
if (classNeighbor.isPresent()) {
    clazz = pool.toClass(newClass, classNeighbor.get());
} else {
    clazz = pool.toClass(newClass);
}
```

then you can provide your own version of `ExpressionCompiler`

petergeneric:
My understanding from reading the javassist javadoc is that the neighbour needs to be in the same package as CtClass, so the choice of neighbour class is tightly coupled to the ExpressionCompiler's choice of package for the CtClass, so this might be too general a solution?

The javadoc on `toClass(Class)` says "Warning: This method should not be used in Java 11 or later. Use toClass(CtClass, Class)." so my statement above might have been too cautious.

[...]

PR #214, opened 2023-07-19
Update ExpressionCompiler to run on JDK17+ without 'add-opens' JVM opt using Javassist ClassPool.toClass overload that takes a neighbour class

petergeneric:
Addresses #160 by pointing Javassist at a neighbour (a class in the same package) class of the class we're generating, allowing it to avoid using the deprecated `ClassLoader.defineClass` (which now requires 'add-opens' to use because Javassist must use reflection to make the method visible), and instead use `MethodHandles`.

This implementation introduces a protected method `instantiateClass` that can be overridden by `ExpressionCompiler` subclassers who are customising the exact mechanics of the class instantiation process if needed.

### DIFF
--- a/src/main/java/ognl/enhance/ExpressionCompiler.java
+++ b/src/main/java/ognl/enhance/ExpressionCompiler.java
@@ -436,7 +436,7 @@ public class ExpressionCompiler implements OgnlExpressionCompiler {
         try {
             newClass.addConstructor(CtNewConstructor.defaultConstructor(newClass));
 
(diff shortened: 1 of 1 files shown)


lukaszlenart:
After merging this change a lot of tests fail - it works good on JDK17. I'm going to revert this PR and introduce this change once OGNL will switch to JDK17

lukaszlenart:
You can always use `OgnlRuntime.setCompiler()` to provide your own version with proper implementation of `instantiateClass()` method.

PR #264, opened 2024-04-18
CONFSRVDEV-29911 update OgnlRuntime & ObjectPropertyAccessor to do access check for set field value

jefferyxhy:
WW-5417

`OgnlRuntime.setFieldValue` doesn't check member access rights via `MemberAccess` interface

 &nbsp;

**Reason**

* Investigation shows that `getMethodValue`/ `setMethodValue` / `getFieldValue` are all updated with member access rights check but not `setFieldValue`, which cause `ObjectPropertyAccessor#setPossibleProperty` expose to security vuln.
* `ObjectPropertyAccessor#setPossibleProperty` has a fallback mechanism using `getWriteMethod` which also lack member access rights check
 
 &nbsp;

**Changes/ Solution**

* add field member access check to `OgnlRuntime#setFieldValue` that is controlled by parameter `checkAccessAndExistence`
* add method member access check to `ObjectPropertyAccessor#setPossibleProperty` code block that uses `OgnlRuntime#getWriteMethod`
 
 &nbsp;

**Result & Impact**
now `ObjectPropertyAccessor#setPossibleProperty` will also check member access rights when fallback to use:
* OgnlRuntime.setFieldValue
* method invoke that is from OgnlRuntime.getWriteMethod

### DIFF
--- a/src/main/java/ognl/ObjectPropertyAccessor.java
+++ b/src/main/java/ognl/ObjectPropertyAccessor.java
@@ -49,7 +49,7 @@ public class ObjectPropertyAccessor implements PropertyAccessor {
 
     /**
      * Returns OgnlRuntime.NotFound if the property does not exist.
-     * 
+     *
      * @param context the current execution context.
      * @param target the object to get the property from.
      * @param name the name of the property to get.
@@ -80,7 +80,7 @@ public class ObjectPropertyAccessor implements PropertyAccessor {
 
     /**
      * Returns OgnlRuntime.NotFound if the property does not exist.
-     * 
+     *
      * @param context the current execution context.
      * @param target the object to set the property in.
      * @param name the name of the property to set.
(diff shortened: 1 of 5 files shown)


src/main/java/ognl/ObjectPropertyAccessor.java
```diff
@@ -97,13 +97,13 @@ public Object setPossibleProperty(Map context, Object target, String name, Objec
         try {
             if (!OgnlRuntime.setMethodValue(ognlContext, target, name, value, true))
             {
-                result = OgnlRuntime.setFieldValue(ognlContext, target, name, value) ? null : OgnlRuntime.NotFound;
+                result = OgnlRuntime.setFieldValue(ognlContext, target, name, value, true) ? null : OgnlRuntime.NotFound;
```
kusalk, line 100: @lukaszlenart This is the bug I reported earlier with respect to public fields

src/main/java/ognl/ObjectPropertyAccessor.java
```diff
@@ -97,13 +97,13 @@ public Object setPossibleProperty(Map context, Object target, String name, Objec
...
             }
 
             if (result == OgnlRuntime.NotFound)
             {
                 Method m = OgnlRuntime.getWriteMethod(target.getClass(), name);
-                if (m != null)
+                if (m != null && ognlContext.getMemberAccess().isAccessible(context, target, m, name))
```
kusalk, line 106: This is a new bug discovered by @jefferyxhy which is much worse, it allows public setters on excluded classes to be called!

kusalk:
Atlassian has already released this patch as `3.3.4-atlassian-1` for immediate integration into our products

lukaszlenart (approved):

lukaszlenart:
I cherry-picked this change into #265

PR #312, opened 2024-11-26
add addDefaultContext in Ognl#getValue to keep same behavior with 3.3.x

Allan-QLB:


lukaszlenart:
As far as I recall, I removed this code on purpose, to lower potential security vulnerabilities. And to keep the evaluation safe, the context must immutable.

PR #313, opened 2024-11-27
add parameter ignoreReadMethod in ObjectPropertyAccessor

Allan-QLB:
current implementation of getting property using `ObjectPropertyAccessor` will firstly try to find get method of target property,
if no get method found then try to find read method. If read method exist returns the return value of the read method。Sometimes read method does not directly returns the value of property. For example   the class `org.apache.kafka.clients.consumer.internals.Fetcher` , it has a property `completedFetches`, and a method `hasCompletedFetches` which will be recognized as a read method of the property `completedFetches`,  it returns a boolean value instead of the actural value of `completedFetches`. But what I need is to get the elements of ConcurrentLinkedQueue. So I think we should offer a parameter to ignore detecting read method in this case.
```
public class Fetcher<K, V> implements Closeable {
   
    private final int maxPollRecords;
    private final boolean checkCrcs;
    private final String clientRackId;
    private final ConsumerMetadata metadata;
    private final FetchManagerMetrics sensors;
    private final SubscriptionState subscriptions;
    private final ConcurrentLinkedQueue<CompletedFetch> completedFetches;
   
    private final ApiVersions apiVersions;

    private PartitionRecords nextInLineRecords = null;

  
    public boolean hasCompletedFetches() {
        return !completedFetches.isEmpty();
    }
}
```

### DIFF
--- a/src/main/java/ognl/ObjectPropertyAccessor.java
+++ b/src/main/java/ognl/ObjectPropertyAccessor.java
@@ -32,20 +32,6 @@ import java.lang.reflect.Method;
 public class ObjectPropertyAccessor implements PropertyAccessor {
 
     /**
-     * Ignore detecting and invoking read method when get property value.
-     */
-    private final boolean ignoreReadMethod;
-
-
-    public ObjectPropertyAccessor() {
-        this(false);
-    }
-
-    public ObjectPropertyAccessor(boolean ignoreReadMethod) {
-        this.ignoreReadMethod = ignoreReadMethod;
-    }
-
-    /**
      * Returns OgnlRuntime.NotFound if the property does not exist.
      *
      * @param context the current execution context.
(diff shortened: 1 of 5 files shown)


src/main/java/ognl/ObjectPropertyAccessor.java
```diff
@@ -31,6 +31,20 @@
...
+     * Ignore detecting and invoking read method when get property value.
+     */
+    private final boolean ignoreReadMethod;
+
+
+    public ObjectPropertyAccessor() {
+        this(false);
+    }
+
+    public ObjectPropertyAccessor(boolean ignoreReadMethod) {
```
lukaszlenart, line 44: It would be good to document this parameter here as well. Right now it's easy to find how to use it, yet users can be confused as this parameter is just passed through to `OgnlRuntime` and the real logic happens there.

Also I wonder if this is a good approach, maybe it should be a Context scoped option, eg.:

```java
OgnlContext context = Ognl.createDefaultContext(null, new ExcludedObjectMemberAccess(true));
context.putIfAbsent("ignore_read_method", Boolean.TRUE);

...

private static boolean shouldGetReadMethod(OgnlContext context) {
  Object value = context.get("ignore_read_method");
  return value instanceof Boolean ? (Boolean) value : true;
}    


if (m == null && shouldGetReadMethod(context))
  m = getReadMethod((target == null) ? null : target.getClass(), propertyName, null);
```

`getMethodValue` is getting to have too many boolean flags

Allan-QLB: ok

lukaszlenart (approved):

KaporMirr (commented):
Na hovno

PR #353, opened 2025-01-23
Reuses ClassResolver and TypeConverter of old OgnlContext

### LINKED ISSUE #350: ognl 3.4.5 - Issue with using custom class loader in getValue
At Apache Camel we have a camel-ognl module
https://github.com/apache/camel/tree/main/components/camel-ognl

When we upgrade from 3.4.4 to 3.4.5 we had an unit test error in one of our tests.
https://github.com/apache/camel/blob/main/components/camel-ognl/src/test/java/org/apache/camel/language/ognl/OgnlTest.java#L48

At first sight I suspect its this change
https://github.com/orphan-oss/ognl/pull/337

in the `getValue` method, that somehow the custom classloader that Camel will provide in the unit test is not being in use.

The custom class loader is part of the test
https://github.com/apache/camel/blob/main/components/camel-ognl/src/test/java/org/apache/camel/language/ognl/MyClassResolver.java#L30

where the classloader will now correctly fix the classname and load the class (notice the Animal1 vs Animal) in the name.

The camel code that creates the ognl context and invokes ognl from camel in that test is executing this code
https://github.com/apache/camel/blob/main/components/camel-ognl/src/main/java/org/apache/camel/language/ognl/OgnlExpression.java#L53

lukaszlenart:
Closes #350

### DIFF
--- a/src/main/java/ognl/Ognl.java
+++ b/src/main/java/ognl/Ognl.java
@@ -268,7 +268,7 @@ public abstract class Ognl {
      * appropriately
      */
     public static OgnlContext addDefaultContext(Object root, OgnlContext context) {
-        return addDefaultContext(root, context.getMemberAccess(), null, null, context);
+        return addDefaultContext(root, context.getMemberAccess(), context.getClassResolver(), context.getTypeConverter(), context);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


davsclaus (approved):

davsclaus:
Thanks for maintaining open source libraries and fixing this

PR #590, opened 2026-06-09
fix: throw OgnlException for out-of-range integer literals

### LINKED ISSUE #583: Ognl.parseExpression throws unchecked NumberFormatException for integer literals exceeding Integer.MAX_VALUE and for all hexadecimal literals
**Repository**: https://github.com/orphan-oss/ognl
**Type**: Bug (parser robustness / API contract violation)
**Affected versions**: 3.3.5, 3.4.7, 3.4.10, 3.4.11, 3.5.0-BETA4 (all tested; presumed older too)
**Severity**: Low (denial-of-service class) — fix is a few-line patch
**Discoverer**: Coverage-guided fuzzing of `ognl.Ognl.parseExpression` with [Jazzer](https://github.com/CodeIntelligenceTesting/jazzer) during 2026-05-29 security review of Apache Struts 7.1.1.

---

## Summary

`Ognl.parseExpression(String)` is documented to throw `ognl.OgnlException` on malformed input, which downstream users (Apache Struts, Apache Camel, Apache Click, Atlassian Confluence pre-7, etc.) `try` / `catch` accordingly.

For certain integer literal inputs, the parser instead lets a raw `java.lang.NumberFormatException` (a `RuntimeException`) escape uncaught. This breaks the API contract — callers that only `catch (OgnlException)` propagate the unexpected exception to their own callers, where it can manifest as HTTP 500 with stack-trace disclosure or as a silently dropped operation.

Affected inputs include:
- Decimal integer literals greater than `Integer.MAX_VALUE` (e.g. `2147483648`, `9999999999`)
- The bare token `-2147483648` (= `Integer.MIN_VALUE`) — the lexer tokenises the minus separately and then tries to parse `2147483648`
- All hexadecimal literals (`0x...`), because `Integer.valueOf(s, 16)` is signed-only
- `Long` literals (`xL` suffix) above `Long.MAX_VALUE`

## Reproduction

```java
import ognl.Ognl;
import ognl.OgnlException;

public class Repro {
    public static void main(String[] args) {
        String[] inputs = {
            "2147483647",                  // OK    (= Integer.MAX_VALUE)
            "2147483648",                  // BUG   (NumberFormatException)
            "-2147483648",                 // BUG   (= Integer.MIN_VALUE, but tokenised as `-` + `2147483648`)
            "9223372036854775807L",        // OK    (= Long.MAX_VALUE)
            "9223372036854775808L",        // BUG
            "0xFFFFFFFF",                  // BUG   (unsigned 32-bit max)
            "0x100000000",                 // BUG
            "0xFFFFFFFFFFFFFFFFL",         // BUG
            "1e100",                       // OK    (handled by makeFloat)
            "999999999999999999.9"         // OK    (handled by makeFloat)
        };
        for (String s : inputs) {
            try {
                Ognl.parseExpression(s);
                System.out.println("OK:               " + s);
            } catch (OgnlException oe) {
                System.out.println("OgnlException:    " + s + " — " + oe.getMessage().split("\n")[0]);
            } catch (Throwable t) {
                System.out.println("UNCAUGHT " + t.getClass().getSimpleName() + ": "
                                    + s + " — " + t.getMessage());
            }
        }
    }
}
```

Run with `ognl-3.4.11.jar` (or any tested version) on the classpath:

```
OK:               2147483647
UNCAUGHT NumberFormatException: 2147483648 — For input string: "2147483648"
UNCAUGHT NumberFormatException: -2147483648 — For input string: "2147483648"
OK:               9223372036854775807L
UNCAUGHT NumberFormatException: 9223372036854775808L — For input string: "9223372036854775808"
UNCAUGHT NumberFormatException: 0xFFFFFFFF — For input string: "FFFFFFFF" under radix 16
UNCAUGHT NumberFormatException: 0x100000000 — For input string: "100000000" under radix 16
UNCAUGHT NumberFormatException: 0xFFFFFFFFFFFFFFFFL — For input string: "FFFFFFFFFFFFFFFF" under radix 16
OK:               1e100
OK:               999999999999999999.9
```

Stack trace (for the `2147483648` case):

```
java.lang.NumberFormatException: For input string: "2147483648"
    at java.base/java.lang.NumberFormatException.forInputString(NumberFormatException.java:67)
    at java.base/java.lang.Integer.parseInt(Integer.java:662)
    at java.base/java.lang.Integer.valueOf(Integer.java:963)
    at ognl.OgnlParserTokenManager.makeInt(OgnlParserTokenManager.java:63)
    at ognl.OgnlParserTokenManager.TokenLexicalActions(OgnlParserTokenManager.java:1659)
    at ognl.OgnlParserTokenManager.getNextToken(OgnlParserTokenManager.java:1532)
    at ognl.OgnlParser.jj_ntk(OgnlParser.java:3155)
    at ognl.OgnlParser.unaryExpression(OgnlParser.java:1082)
    ... (parser chain)
    at ognl.Ognl.parseExpression(Ognl.java:179)
```

## Affected versions matrix

| Version | Released | Reproducible |
|---------|----------|:------------:|
| 3.3.5 | 2024 | ✓ |
| 3.4.7 | 2024 | ✓ |
| 3.4.10 | 2025 | ✓ |
| 3.4.11 | 2025 | ✓ |
| 3.5.0-BETA4 | 2025 | ✓ |

Confirmed against `maven-metadata.xml`'s `latest = 3.5.0-BETA4`.

## Root cause

In `src/main/java/ognl/OgnlParserTokenManager.java` (generated from `ognl.jj` / `ognl.jjt`):

```java
private Object makeInt()
{
    Object  result;
    String  s = image.toString();
    int     base = 10;

    if ( s.charAt(0) == '0' )
        base = (s.length() > 1 && (s.charAt(1) == 'x' || s.charAt(1) == 'X'))? 16 : 8;
    if ( base == 16 )
        s = s.substring(2); // Trim the 0x off the front
    switch ( s.charAt(s.length()-1) ) {
        case 'l': case 'L':
            result = Long.valueOf( s.substring(0,s.length()-1), base );      // ← NumberFormatException leaks
            break;

        case 'h': case 'H':
            result = new BigInteger( s.substring(0,s.length()-1), base );    // (no overflow issue)
            break;

        default:
            result = Integer.valueOf( s, base );                              // ← NumberFormatException leaks
    }
    return result;
}
```

`Integer.valueOf(s, base)` and `Long.valueOf(s, base)` throw `NumberFormatException` (`RuntimeException`) when the value is out of range. The lexer action returns to JavaCC-generated code that does not wrap the exception, and it propagates all the way to `Ognl.parseExpression`'s caller.

The `-` minus-sign case (`-2147483648`) is a second sub-bug: the JavaCC grammar tokenises the leading minus as a separate `UNARY_MINUS` token, so the integer literal `makeInt` sees is the bare positive token `2147483648`, which is exactly one over `Integer.MAX_VALUE`. The fix below resolves both.

## Suggested fix

Replace the `default:` (and optionally the `L`-suffix) branch with a graceful widening that always returns *some* numeric value:

```java
default:
    try {
        result = Integer.valueOf( s, base );
    } catch (NumberFormatException nfeInt) {
        try {
            result = Long.valueOf( s, base );
        } catch (NumberFormatException nfeLong) {
            // Promote beyond long to BigInteger so the parser never throws raw NFE.
            result = new BigInteger( s, base );
        }
    }
    break;
```

And similarly for the `L`-suffix branch:

```java
case 'l': case 'L':
    try {
        result = Long.valueOf( s.substring(0,s.length()-1), base );
    } catch (NumberFormatException nfe) {
        // L was explicit but value overflows long; widen to BigInteger so we
        // surface a TokenMgrError / OgnlException up the stack instead of NFE.
        throw new TokenMgrError(
            "Integer literal with 'L' suffix exceeds Long.MAX_VALUE: " + s,
            TokenMgrError.LEXICAL_ERROR);
    }
    break;
```

For the `default:` branch the silent widening keeps existing scripts that use unsuffixed numerics ≤ Long.MAX_VALUE working; only callers that were silently coerced into `Integer` see a behaviour change (now they get `Long` or `BigInteger`). For the `L`-explicit branch, raising `TokenMgrError` is the consistent choice because the user intent was a `long`-typed value.

If a fully backwards-compatible patch is preferred, replace **only** the raw `NumberFormatException` cases with `TokenMgrError` (so callers always see `OgnlException` regardless of input):

```java
default:
    try {
        result = Integer.valueOf( s, base );
    } catch (NumberFormatException nfe) {
        throw new TokenMgrError(
            "Integer literal out of range for Integer (use Long with 'L' suffix or use BigInteger with 'H' suffix): "
            + s, TokenMgrError.LEXICAL_ERROR);
    }
    break;
```

The minus-sign case (`-2147483648`) is then surfaced by the parser as an `OgnlException` at the `UNARY_MINUS expression` reduction — acceptable, since the user can write `-9999999999L` instead and existing scripts that relied on this exact corner case are vanishingly rare.

## Test cases to add

```
src/test/java/ognl/OgnlIntegerLiteralBoundaryTest.java
```

```java
@Test public void intMaxOk() throws Exception { Ognl.parseExpression("2147483647"); }
@Test(expected = OgnlException.class) public void intMaxPlusOneRaisesOgnlException() throws Exception { Ognl.parseExpression("2147483648"); }
@Test(expected = OgnlException.class) public void hexUnsignedMaxRaisesOgnlException() throws Exception { Ognl.parseExpression("0xFFFFFFFF"); }
@Test public void longMaxOk() throws Exception { Ognl.parseExpression("9223372036854775807L"); }
@Test(expected = OgnlException.class) public void longMaxPlusOneRaisesOgnlException() throws Exception { Ognl.parseExpression("9223372036854775808L"); }
@Test(expected = OgnlException.class) public void intMinTokenRaisesOgnlException() throws Exception { Ognl.parseExpression("-2147483648"); }
```

Or, if going with the silent-widening fix, change the last test class names to `…WidensTo…` and assert the parsed tree's numeric class.

## Downstream impact

I discovered this while fuzzing the `Apache Struts 7.1.1` OGNL surface. Struts mitigates the user-facing impact in [`ParametersInterceptor.applyParametersOnStack`](https://github.com/apache/struts/blob/STRUTS_7_1_1/core/src/main/java/org/apache/struts2/interceptor/parameter/ParametersInterceptor.java) by wrapping `stack.setParameter(...)` in `catch (RuntimeException)` (silent drop in production, dev-mode log). So no remote crash there.

However, the `StrutsRequestWrapper.getAttribute → ValueStack.findValue` path (reachable from Tiles iteration of request attribute names) does **not** wrap `RuntimeException`. An attacker that can cause `getAttribute("2147483648")` to be invoked will get the raw NFE bubbled to the servlet container, which may surface as HTTP 500 with stack trace disclosure depending on container config. Practical exploitability is low because the JSP/JSTL author normally chooses the attribute key, but worth noting.

Other downstream users of `Ognl.parseExpression` who don't catch `RuntimeException` (most code, by default) are exposed identically.

## How to reproduce my fuzz finding

Compile this target against `ognl-3.4.11.jar` and `jazzer_standalone.jar`:

```java
import com.code_intelligence.jazzer.api.FuzzedDataProvider;
public class OgnlFuzzer {
    public static void fuzzerTestOneInput(FuzzedDataProvider data) {
        String expr = data.consumeRemainingAsString();
        if (expr == null || expr.isEmpty()) return;
        try { ognl.Ognl.parseExpression(expr); }
        catch (ognl.OgnlException oe) { /* expected */ }
        catch (Throwable t) { throw new RuntimeException("Unexpected parser throwable", t); }
    }
}
```

Run with libFuzzer args:

```
jazzer --cp=ognl-3.4.11.jar:. --target_class=OgnlFuzzer -max_total_time=600 corpus/
```

A 15-byte crash `4444444444E\x00\x00\x000` is typically found in under 5 minutes. The crash minimises to `2147483648`.

## Suggested follow-ups for the project

1. Apply the patch above (silent-widening or `TokenMgrError`; my preference is silent-widening for backwards compatibility on the unsuffixed `default` branch).
2. Add the boundary test class.
3. Run a short Jazzer campaign in CI (≥ 10 minutes) against `Ognl.parseExpression` to catch future regressions.
4. Consider documenting in `Ognl.parseExpression`'s javadoc that the method may throw `OgnlException` for inputs that exceed numeric type ranges, so callers know to expect the wrapped form.

I'd be happy to send a PR with the patch + tests if a maintainer confirms which fix flavour you prefer.

---

**Reporter contact**: Yukihiro Nakamura <nakamurabeer@gmail.com>
**Disclosure date**: 2026-05-29
**Embargo**: None requested (bug is denial-of-service class with low real-world exploitability; coordinated disclosure not necessary, but I'm holding off on a public-blog write-up until the project has a chance to respond.)

SAY-5:
Fixes #583. Integer literals that overflow their target type (decimal above Integer.MAX_VALUE, long above Long.MAX_VALUE, and all hex literals in the upper half of the unsigned range) currently let a raw NumberFormatException escape parseExpression, which breaks the documented OgnlException contract that callers like Struts rely on. This wraps the failure in makeInt() as a TokenMgrError so it surfaces through the existing parseExpression handler as an OgnlException. Added a parameterized test covering the decimal, long, and hex overflow cases.

### DIFF
--- a/ognl/src/main/java/ognl/Ognl.java
+++ b/ognl/src/main/java/ognl/Ognl.java
@@ -115,7 +115,8 @@ public abstract class Ognl {
      *
      * @param expression the OGNL expression to be parsed
      * @return a tree representation of the expression
-     * @throws ExpressionSyntaxException if the expression is malformed
+     * @throws ExpressionSyntaxException if the expression is malformed, including an integer
+     *                                   literal whose value is outside the representable range
      * @throws OgnlException             if there is a pathological environmental problem
      */
     public static Object parseExpression(String expression) throws OgnlException {
--- a/ognl/src/main/javacc/ognl.jj
+++ b/ognl/src/main/javacc/ognl.jj
@@ -1916,7 +1916,10 @@ TOKEN_MGR_DECLS:
         } catch ( NumberFormatException e ) {
             // Surface as a lexical error so callers see the documented OgnlException
             // instead of a raw NumberFormatException (e.g. literals out of range).
-            throw new TokenMgrError( "Integer literal out of range: " + image, TokenMgrError.LEXICAL_ERROR );
+            // TokenMgrError has no cause-accepting constructor, so fold the original
+            // detail into the message to keep diagnostics and use the caught exception.
+            throw new TokenMgrError( "Integer literal out of range: " + image
+                    + " (" + e.getMessage() + ")", TokenMgrError.LEXICAL_ERROR );
         }
         return result;
     }
--- a/ognl/src/main/javacc/ognl.jj
+++ b/ognl/src/main/javacc/ognl.jj
@@ -1899,18 +1899,24 @@ TOKEN_MGR_DECLS:
             base = (s.length() > 1 && (s.charAt(1) == 'x' || s.charAt(1) == 'X'))? 16 : 8;
         if ( base == 16 )
             s = s.substring(2); // Trim the 0x off the front
-        switch ( s.charAt(s.length()-1) ) {
-            case 'l': case 'L':
-                result = Long.valueOf( s.substring(0,s.length()-1), base );
-                break;
+        try {
+            switch ( s.charAt(s.length()-1) ) {
+                case 'l': case 'L':
+                    result = Long.valueOf( s.substring(0,s.length()-1), base );
+                    break;
 
-            case 'h': case 'H':
-                result = new BigInteger( s.substring(0,s.length()-1), base );
-                break;
+                case 'h': case 'H':
+                    result = new BigInteger( s.substring(0,s.length()-1), base );
+                    break;
 
-            default:
-                result = Integer.valueOf( s, base );
-                break;
+                default:
+                    result = Integer.valueOf( s, base );
+                    break;
+            }
+        } catch ( NumberFormatException e ) {
+            // Surface as a lexical error so callers see the documented OgnlException
+            // instead of a raw NumberFormatException (e.g. literals out of range).
+            throw new TokenMgrError( "Integer literal out of range: " + image, TokenMgrError.LEXICAL_ERROR );
         }
         return result;
     }
--- a/ognl/src/test/java/ognl/test/IntegerLiteralOverflowTest.java
+++ b/ognl/src/test/java/ognl/test/IntegerLiteralOverflowTest.java
@@ -20,10 +20,10 @@ package ognl.test;
 
 import ognl.Ognl;
 import ognl.OgnlException;
-import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.ValueSource;
 
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 
 class IntegerLiteralOverflowTest {
@@ -31,6 +31,8 @@ class IntegerLiteralOverflowTest {
     @ParameterizedTest
     @ValueSource(strings = {
             "2147483648",            // decimal above Integer.MAX_VALUE
+            "-2147483648",           // lexer tokenises the leading '-' separately,
+                                     // so makeInt() sees the bare 2147483648
             "9999999999",            // larger decimal
             "9223372036854775808L",  // long above Long.MAX_VALUE
             "0xFFFFFFFF",            // hex above signed 32-bit range
@@ -39,5 +41,16 @@ class IntegerLiteralOverflowTest {
     })
     void shouldThrowOgnlExceptionForOutOfRangeIntegerLiteral(String expression) {
         assertThrows(OgnlException.class, () -> Ognl.parseExpression(expression));
+    }
+
+    @ParameterizedTest
+    @ValueSource(strings = {
+            "2147483647",            // Integer.MAX_VALUE
+            "9223372036854775807L",  // Long.MAX_VALUE
+            "0x7FFFFFFF",            // largest in-range hex int
+            "1e100"                  // float path is unaffected by makeInt()
+    })
+    void shouldParseInRangeNumericLiteral(String expression) {
+        assertDoesNotThrow(() -> Ognl.parseExpression(expression));
     }
 }
--- a/ognl/src/test/java/ognl/test/IntegerLiteralOverflowTest.java
+++ b/ognl/src/test/java/ognl/test/IntegerLiteralOverflowTest.java
@@ -1,0 +1,43 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package ognl.test;
+
+import ognl.Ognl;
+import ognl.OgnlException;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
+
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+class IntegerLiteralOverflowTest {
+
+    @ParameterizedTest
+    @ValueSource(strings = {
+            "2147483648",            // decimal above Integer.MAX_VALUE
+            "9999999999",            // larger decimal
+            "9223372036854775808L",  // long above Long.MAX_VALUE
+            "0xFFFFFFFF",            // hex above signed 32-bit range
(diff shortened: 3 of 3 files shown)


lukaszlenart (commented):
Thanks for this, @SAY-5 — the core approach is spot on. Wrapping the `switch` in `makeInt()` and rethrowing `NumberFormatException` as a `TokenMgrError(LEXICAL_ERROR)` is exactly the right fix, since `parseExpression` already converts `TokenMgrError` into an `OgnlException`. A few suggestions before merge:

**1. SonarCloud (`java:S1166`) — preserve/use the caught exception.** The `catch (NumberFormatException e)` never references `e`, and `TokenMgrError` has no cause-accepting constructor, so the original detail is dropped. Our quality gate will likely flag this on new code. Folding the cause into the message resolves it and keeps useful diagnostics:

```java
} catch (NumberFormatException e) {
    // Surface as a lexical error so callers see the documented OgnlException
    // instead of a raw NumberFormatException. TokenMgrError has no
    // cause-accepting constructor, so the original message is folded in.
    throw new TokenMgrError("Integer literal out of range: " + image
            + " (" + e.getMessage() + ")", TokenMgrError.LEXICAL_ERROR);
}
```

**2. Add the `-2147483648` case.** The issue calls this out specifically: the lexer tokenises the leading `-` separately, so `makeInt()` sees the bare `2147483648`. It's worth a test to lock in that it surfaces as `OgnlException`.

**3. Assert the valid boundaries still parse.** A couple of "does not throw" cases (`2147483647`, `9223372036854775807L`) guard against an over-eager future change rejecting in-range values.

**(Optional) 4.** A note on `parseExpression`'s javadoc that it throws `OgnlException` for out-of-range numeric literals, and a quick confirmation test that the float path (`1e100`) is unaffected, would round it out nicely.

The parameterized test reads well and the `ognl.test` package placement is consistent with the rest of the suite. With the S1166 tweak and the `-2147483648` case, this is good to go. Thanks again!

lukaszlenart (approved):
Thanks, LGTM!

